Issue #1, opened 2020-03-04
Any plans to cache results?


cbuijs:
Hi,

Are there any plans to add caching of results that were fetched?

-Chris

folbricht:
That's not actually hard to do. I can add a cache object similar to a group. That way it's possible to have multiple caches, or just cache certain upstream resolvers. Going to take a look at adding this.

cbuijs:
@folbricht 

That would be even more then I could hoped for :-)

-Chris

folbricht:
First attempt in https://github.com/folbricht/routedns/pull/3, just need to do a bit more testing. Feel free to try it out as well.

cbuijs:
@folbricht 

Just did a first round of testing and seems to be working fine and quite fast. Things to consider:

- Minimum TTL: This would be the minimal amount of time a record would stay in cache, if a TTL of a record coming back is below this, it will be set to this value (Defaul would be "0').
- Maximum TTL:
- Negative TTL: Specify caching TTL of records with RCodes other then NOERROR.
- Blacklist TTL: Set TTL for blacklisted records (might be hard as is other object, not a biggie).

-Chris

folbricht:
I couldn't find a reference implementation so this is what the cache object is doing at the moment. Before an answer goes into the cache, I go through all answer, NS, extra records to find the lowest TTL. The lowest TTL then becomes the lifetime of the answer in the cache. On the next query when an answer is found in the cache, it'll go through all records again and subtract the time it spent in the cache. So if the item was added 5 seconds ago, all records will have their TTL reduced by 5 before the answer is returned. TTL is a uint32 so it can't go negative. 

Do you see any issues with that logic?

A couple more things I need to add:
- Don't cache error responses
- Cache NXDOMAIN responses if there's a SOA record (using the TTL of the SOA)

cbuijs:
No issues with the logic, but... 

CDN's tend use long CNAME-chains with differnt TTLs per record, where the A/AAAA record(s) they point to are actually leading and used if you think about it. I would take the (lowest) TTL of those and generalize it for the whole RRSET (e.g. use same TTL for all records in the RRSET). This is my personal preference and never encountered any issues. It is a gray area for sure. Maybe another configuration option? :-).

[...]

PR #3, opened 2020-04-02
New cache group

folbricht:
Implements a cache element, as per https://github.com/folbricht/routedns/issues/1


folbricht:
Needs to handle negative responses as per https://tools.ietf.org/html/rfc2308

Issue #5, opened 2020-04-03
Allow an "include" statement to refer to another toml file for config to be included


cbuijs:
Hi,

Subject says it all basically.

Basically if you want to generate/automate toml files it becomes handy in auto-config environment. Pointing it to a "conf.d" directory and read all toml files from there would be paradise on this :-).

Cheers,
-Chris.

folbricht:
How about allowing multiple files on the command-line? Could look like:
```
routedns /etc/routedns/conf.d/*.conf /some/other/file.conf
```

cbuijs:
@folbricht 

Elegant alternative!

folbricht:
Done in https://github.com/folbricht/routedns/pull/6, give it a try. All it does is to concatenate the different files and then load it whole. See https://github.com/folbricht/routedns/tree/master/cmd/routedns/example-config/split-config for an exmple

cbuijs:
Works great!

PR #6, opened 2020-04-04
Support configuration split over multiple files

folbricht:
Allows multiple configuration files as per #5

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,12 @@ The config file defines listeners which represent open ports and protocols on wh
 
 ![configuration](doc/configuration.png)
 
+Configuration files can be broken up into individual files to support large or generated configurations. Split configuration files are passed as arguments:
+
+```text
+routedns example-config/split-config/*.toml
+```
+
 ### Resolvers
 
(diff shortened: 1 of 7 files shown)


cbuijs:
Just checked it out, and found one "shortcoming":

You cannot have the same sections in different files.

For example, I have a `[groups]` section in my first toml file for generic grouping, and want to have another `[groups]` (blocklist) in another toml-file, when done, the below error pops up, so I guess consolation is indeed a (straight) concatenation as said earlier, not a "merge" of sections.

The error:
`Error: Near line 307 (last key parsed ''): Key 'groups' has already been defined.`

I point the command-line to `/etc/routedns/conf.d/*.toml`

I have two toml files in there:
- `00-config.toml`
- `10-blocklist.toml`

The `00-config.toml` has my standard stuff, the `10-blocklist.toml` has only the `[groups]` section, with a `[groups.blocklist]` entry with some domain blocking in there and a default resolver (which is in `00-config.toml`). If I merge everything in one file, all works fine.

Pretty please (merge sections from all toml files instead of consolication)? :-)

folbricht:
I should have said, you need to leave out the `[groups]`, `[resolvers]`section in the config files. Those markers aren't actually required anyway and just there for easier reading of large files. Can you try that? If that doesn't work well enough, I can change it to a merge.

See an example here: https://github.com/folbricht/routedns/tree/master/cmd/routedns/example-config/split-config Going to add some more files there and update docs to that effect.

Issue #64, opened 2020-06-15
Panic when using blocklist-source with blocklist-v2


sandler31:
It seems that the NewDomainDB method does not pass back the loader it was given (seen [here](https://github.com/folbricht/routedns/blob/master/blocklistdb-domain.go#L58)).
Thus, when update is made and the following code is executed:
```golang
func (m *DomainDB) Reload() (BlocklistDB, error) {
	return NewDomainDB(m.loader)
}
```
a panic follows.

Relevant panic:
```
goroutine 35 [running]:
github.com/folbricht/routedns.NewDomainDB(0x0, 0x0, 0xb75dd0, 0xe, 0xc000245d00)
	/home/mich/data/proj/dwall/vendor/github.com/folbricht/routedns/blocklistdb-domain.go:28 +0xbf
github.com/folbricht/routedns.(*DomainDB).Reload(0xc00029c0c0, 0x0, 0x0, 0x0, 0x0)
	/home/mich/data/proj/dwall/vendor/github.com/folbricht/routedns/blocklistdb-domain.go:64 +0xb1
github.com/folbricht/routedns.MultiDB.Reload(0xc00029c1e0, 0x2, 0x2, 0x0, 0x0, 0x0, 0x0)
	/home/mich/data/proj/dwall/vendor/github.com/folbricht/routedns/blocklistdb-multi.go:24 +0x89
github.com/folbricht/routedns.(*Blocklist).refreshLoopAllowlist(0xc00028e240, 0x12a05f200)
	/home/mich/data/proj/dwall/vendor/github.com/folbricht/routedns/blocklist.go:164 +0x118
created by github.com/folbricht/routedns.NewBlocklist
	/home/mich/data/proj/dwall/vendor/github.com/folbricht/routedns/blocklist.go:54 +0xf9
```

My configuration:
```
[resolvers.cloudflare-doh]
address = "https://1.1.1.1/dns-query{?dns}"
protocol = "doh"

[groups.whitelist]
type             = "blocklist-v2"
resolvers        = ["cloudflare-doh"]
blocklist-format = "regexp"            
blocklist        = ['.+']
allowlist-refresh = 5
allowlist-source = [
  {format = "domain", source = "http://internal.server/routedns/allow/domain"},
  {format = "regexp", source = "http://internal.server/routedns/allow/regexp"},
]

[listeners.local-udp]
address = ":5353"
protocol = "udp"
resolver = "whitelist"
```

folbricht:
[...]

PR #65, opened 2020-06-15
Pass back the loader given upon instantiation of the domain BlocklistDB

sandler31:
Resolves issue #64,
Hope it's ok to pull request, didn't see any not by you :)


folbricht:
Whoops, nice find. Thanks for the PR. The CIDR and Location DBs have the same issue. I'll fix those too.

PR #77, opened 2020-07-25
Validate resolver addresses during startup

### LINKED ISSUE #76: IPv6 address need check. Should throw exception, if format illegal.
If IP format is illegal should throw exception.

Illegal example like this:
  [resolvers.google]
  address="2001:4860:4860::8888:53"
  protocol="tcp"

Correct format like this:
address="[2001:4860:4860::8888]:53"

I do not know whether IPv4 address has the same problem.

folbricht:
Fixes #76 by checking resolver endpoint at startup.

### DIFF
--- a/cmd/routedns/main.go
+++ b/cmd/routedns/main.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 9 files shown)


folbricht:
I should really break out the validation piece into it's own library so that can be used independently.

folbricht:
Actually, this will need more work since for DoH, the address isn't just <hostname>:<port>. I'll have to revisit this

PR #78, opened 2020-07-26
Support X-Forwarded-For when it comes from a trusted proxy

SeanBurford:
I find this useful for running routedns behind nginx.

### DIFF
--- a/cmd/routedns/main.go
+++ b/cmd/routedns/main.go
@@ -244,15 +244,15 @@ func start(opt options, args []string) error {
 			if err != nil {
 				return err
 			}
-			proxyAddr := net.ParseIP(l.DoH.ProxyAddr)
-			if l.DoH.ProxyAddr != "" && proxyAddr == nil {
-				return fmt.Errorf("listener '%s' proxy-address '%s' is not an IPv4 or IPv6 address", id, l.DoH.ProxyAddr)
+			httpProxyAddr := net.ParseIP(l.DoH.HTTPProxyAddr)
+			if l.DoH.HTTPProxyAddr != "" && httpProxyAddr == nil {
+				return fmt.Errorf("listener '%s' proxy-address '%s' is not an IPv4 or IPv6 address", id, l.DoH.HTTPProxyAddr)
 			}
 			opt := rdns.DoHListenerOptions{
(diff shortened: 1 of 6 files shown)


cbuijs:
Maybe use same approach as with EDNS (like for subnet caching), and support
for CIDR addresses to designated trusted proxies/subnets?

Not sure about the "proxy" naming convention as it is more http related
then DNS but from the usage/application it makes sense.

SeanBurford:
proxy-subnet is a good idea, however I would prefer to push that back to a future PR because the config parsing will add complexity and:

- Config (trusted-proxy-subnet4, trusted-proxy-subnet6) can be added in a backwards compatible way (default to a subnet of all ones, which is the current behaviour).
- Implementation is straightforward (Change ip.Equal to subnet.Contains).

cmd/routedns/main.go
```diff
@@ -244,7 +244,17 @@ func start(opt options, args []string) error {
...
-			ln, err := rdns.NewDoHListener(id, l.Address, rdns.DoHListenerOptions{TLSConfig: tlsConfig, ListenOptions: opt, Transport: l.Transport}, resolver)
+			proxyAddr := net.ParseIP(l.DoH.ProxyAddr)
+			if l.DoH.ProxyAddr != "" && proxyAddr == nil {
+				return fmt.Errorf("listener '%s' proxy-address '%s' is not an IPv4 or IPv6 address", id, l.DoH.ProxyAddr)
+			}
+			opt := rdns.DoHListenerOptions{
+				TLSConfig:     tlsConfig,
+				ListenOptions: opt,
+				Transport:     l.Transport,
+				ProxyAddr:     &proxyAddr,
```
folbricht, line 255: Do we really need to pass a pointer here? net.IP is already a slice and I believe it would work without it

SeanBurford: Thanks, I've removed the pointer.

SeanBurford:
> Not sure about the "proxy" naming convention as it is more http related then DNS but from the usage/application it makes sense.

I've moved to HttpProxy to make the intention clearer.  Since these are listeners the HTTPProxy must be incoming rather than outbound.

folbricht:
Thanks for the contribution!

PR #79, opened 2020-07-29
Subnet

SeanBurford:
Changes:

- Change from one proxy address to CIDR as previously promised.  
- Fix the docs for the configuration file.
- Change the config section from "doh = { trusted-proxy = ... }" to "frontend = { trusted-proxy = ... }"

This is a breaking change.  Anyone who worked out how to set this up despite the problem with the docs will now fail with the message "invalid CIDR address" because the mask length is now required.  I considered making it optional but decided to go for simplicity (since I'm likely the only user right now).

### DIFF
--- a/cmd/routedns/main.go
+++ b/cmd/routedns/main.go
@@ -244,15 +244,15 @@ func start(opt options, args []string) error {
 			if err != nil {
 				return err
 			}
-			httpProxyAddr := net.ParseIP(l.DoH.HTTPProxyAddr)
-			if l.DoH.HTTPProxyAddr != "" && httpProxyAddr == nil {
-				return fmt.Errorf("listener '%s' proxy-address '%s' is not an IPv4 or IPv6 address", id, l.DoH.HTTPProxyAddr)
+			_, httpProxyNet, err := net.ParseCIDR(l.Frontend.HTTPProxyNet)
+			if l.Frontend.HTTPProxyNet != "" && httpProxyNet == nil {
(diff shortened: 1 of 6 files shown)


cmd/routedns/main.go
```diff
@@ -244,15 +244,15 @@ func start(opt options, args []string) error {
 			if err != nil {
 				return err
 			}
-			httpProxyAddr := net.ParseIP(l.DoH.HTTPProxyAddr)
-			if l.DoH.HTTPProxyAddr != "" && httpProxyAddr == nil {
-				return fmt.Errorf("listener '%s' proxy-address '%s' is not an IPv4 or IPv6 address", id, l.DoH.HTTPProxyAddr)
+			_, httpProxyNet, err := net.ParseCIDR(l.Frontend.HTTPProxyNet)
+			if l.Frontend.HTTPProxyNet != "" && httpProxyNet == nil {
+				return fmt.Errorf("listener '%s' trusted-proxy '%s': %v", id, l.Frontend.HTTPProxyNet, err)
```
folbricht, line 249: While this works as expected, it's a little hard to read due the unusual error handling. How about something like this, a little more explicit:

```go
	var httpProxyNet *net.IPNet
	if l.Frontend.HTTPProxyNet != "" {
		_, httpProxyNet, err = net.ParseCIDR(l.Frontend.HTTPProxyNet)
		if err != nil {
			return fmt.Errorf("listener '%s' trusted-proxy '%s': %v", id, l.Frontend.HTTPProxyNet, err)
		}
	}
```

folbricht:
Thank you

Issue #80, opened 2020-07-30
Metrics


SeanBurford:
With metrics you could graph RouteDNS performance (eg. in Grafana).

The benefits are:

- Less need for logging.  Instead of running at "-l 5" you can see behaviour using counters.
- Ability to add graphing and monitoring.
- Additional insight into memory allocations and usage.

To demonstrate the idea I've created https://github.com/SeanBurford/routedns/commit/3ce505c3b3e18c3e7eddd3db6e4646fded579830

Looking at the golang expvar module, I was concerned that metrics would be exported to everyone who can access the HTTP server (eg. the DoH port).  This turns out to not be the case, since you specify a http mux you have to explicitely add the expvar handler to a path of your choosing.  If the DoHListener config included an optional "vars_path" item users could choose whether to export variables and which path to serve them on.

The output from https://server/routedns/vars which looks like this:

```{"cmdline": ["routedns","config.toml"],
"memstats": ...
"routedns.client.dns.error": {"post": 31},
"routedns.client.dns.query": 146,
"routedns.client.dns.response": {"NOERROR": 114, "NXDOMAIN": 1},
"routedns.client.local-udp.query": 61,
"routedns.client.local-udp.response": {"NOERROR": 46, "NXDOMAIN": 7},
"routedns.listener.srv-doh.drop": 0,
"routedns.listener.srv-doh.error": {},
"routedns.listener.srv-doh.method": {"GET": 1},
"routedns.listener.srv-doh.query": 1,
"routedns.listener.srv-doh.response": {"NXDOMAIN": 1},
"routedns.listener.srv-doh.session": {"tcp": 1},
"routedns.listener.srv-tcp.drop": 0,
"routedns.listener.srv-tcp.header": {},
"routedns.listener.srv-tcp.query": 1,
"routedns.listener.srv-tcp.response": {"NOERROR": 1},
"routedns.listener.srv-tcp.session": {"tcp": 1},
"routedns.listener.srv-udp.drop": 0,
"routedns.listener.srv-udp.header": {"resolve": 7},
"routedns.listener.srv-udp.query": 173,
"routedns.listener.srv-udp.response": {"NOERROR": 159, "NXDOMAIN": 7, "SERVFAIL": 7},
"routedns.listener.srv-udp.session": {"udp": 173},
"routedns.pipeline.local-udp.error": {"idle_timeout": 9, "querytimeout": 8},
"routedns.pipeline.local-udp.maxqlen": 13,
[...]

PR #84, opened 2020-08-10
Expvar

SeanBurford:
Add metrics (including a new "admin" listener).

Add expvar metrics to listeners, routers and clients.  Variables are kept in structs which are reused where possible (eg. routers use or subclass RouterMetrics).

Metrics examples:
"routedns.listener.local-udp.query": 8,
"routedns.listener.local-udp.response": {"NOERROR": 8},
"routedns.router.cloudflare.available": 9,
"routedns.router.cloudflare.route": {"cloudflare-dot-1-0-0-1": 1, "cloudflare-dot-1-1-1-1": 1},
"routedns.client.cloudflare-dot-1-0-0-1.maxqueue": 1,
"routedns.client.cloudflare-dot-1-0-0-1.query": 1,
"routedns.client.cloudflare-dot-1-0-0-1.response": {"NOERROR": 1},

AdminListener is a new listener based on the DoH listener, including accepting the same config options.  This listener is a bit special because it's the only listener that doesn't route requests.  The config for AdminListener has been added to the docs and examples.

### DIFF
--- a/pipeline.go
+++ b/pipeline.go
@@ -16,7 +16,7 @@ const queryTimeout = time.Second
 // Tear down an upstream connection if nothing has been received for this long.
 const idleTimeout = 10 * time.Second
 
-// Pipeline is a DNS client that is able to use pipelining for ultiple requests over
+// Pipeline is a DNS client that is able to use pipelining for multiple requests over
 // one connection, handle out-of-order responses and deals with disconnects
 // gracefully. It opens a single connection on demand and uses it for all queries.
 // It can manage UDP, TCP, DNS-over-TLS, and DNS-over-DTLS connections.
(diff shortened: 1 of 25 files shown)


SeanBurford:
https://github.com/folbricht/routedns/issues/80

pipeline.go
```diff
@@ -142,10 +151,16 @@ func (c *Pipeline) start() {
 				}
 				req := inFlight.get(a) // match the answer to an in-flight query
 				if req == nil {
+					c.metrics.err.Add("unexpected_a", 1)
 					log.WithField("qname", qName(a)).Warn("unexpected answer received, ignoring")
 					continue
 				}
+				c.metrics.response.Add(rCode(a), 1)
 				req.markDone(a, nil)
+				ql := (int64)(inFlight.maxLen)
```
folbricht, line 160: Isn't this a possible race since maxLen isn't locking the mutex in the inFlightQueue? Might be better to add a safe method to the queue to return that value.

folbricht: Addressed in a subsequent change.

folbricht:
Thank you

PR #90, opened 2020-09-23
Store empty answers in cache

### LINKED ISSUE #89: 0-length answers are not cached
When a 0-length response is received, it is not stored in cache. I expect that it should be stored for `cache-negative-ttl` time.
Below is the relevant gdb session part:

```
153	func (r *Cache) storeInCache(query, answer *dns.Msg) {
(gdb) p *answer 
$3 = {MsgHdr = {Id = 34605, Response = true, Opcode = 0, Authoritative = false, Truncated = false, 
    RecursionDesired = true, RecursionAvailable = true, Zero = false, AuthenticatedData = false, 
    CheckingDisabled = false, Rcode = 0}, Compress = false, Question = {array = 0xc0004a00a0, len = 1, cap = 1}, 
  Answer = {array = 0x0, len = 0, cap = 0}, Ns = {array = 0x0, len = 0, cap = 0}, Extra = {array = 0x0, len = 0, 
    cap = 0}}
(gdb) n
154		now := time.Now()
(gdb) 
157		item := &cacheAnswer{Msg: answer, timestamp: now}
(gdb) 
160		min, ok := minTTL(answer)
(gdb) 
163		switch answer.Rcode {
(gdb) n
170		case dns.RcodeSuccess:
(gdb) 
171			if !ok {
(gdb) 
172				return
(gdb) 
```

As you can see, `ok` is `false` in this case, and `storeInCache` returns immediately. It should use `r.NegativeTTL` value instead.

mikhirev:
A 0-length answer has a dns.RcodeSuccess status and no TTL values.
Store it in cache like any other negative answer.

Fixes #89.

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -161,17 +161,12 @@ func (r *Cache) storeInCache(query, answer *dns.Msg) {
 
 	// Calculate expiry for the whole record. Negative answers may not have a SOA to use the TTL from.
 	switch answer.Rcode {
-	case dns.RcodeNameError, dns.RcodeRefused, dns.RcodeNotImplemented, dns.RcodeFormatError:
+	case dns.RcodeSuccess, dns.RcodeNameError, dns.RcodeRefused, dns.RcodeNotImplemented, dns.RcodeFormatError:
 		if ok {
 			item.expiry = now.Add(time.Duration(min) * time.Second)
 		} else {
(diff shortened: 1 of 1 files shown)


folbricht (approved):

folbricht:
Thanks for the contribution!

Issue #91, opened 2020-09-24
Option to enable `SERVFAIL` answers caching


mikhirev:
According to RFC2308,

>   a resolver MAY cache a server failure response.  If it
>   does so it MUST NOT cache it for longer than five (5) minutes, and it
>   MUST be cached against the specific query tuple <query name, type,
>   class, server IP address>.

It would be nice to have an option to enable such caching, e. g. `cache-servfail-ttl`, similar to `cache-negative-ttl`.

folbricht:
Given that negative-ttl is supposed to be quite short anyway, perhaps the servfail could just be cached by default. An option is fine too, but if there're no real downsides, doing it always seems reasonable.

cbuijs:
+1 on re-using `cache-negative-ttl`. Maybe cap that one anyway with 300 secs as max allowed value?

mikhirev:
Well, using value from `cache-negative-ttl`, but not longer than 300 s, sounds good for me.

PR #98, opened 2020-10-23
Cache SERVFAIL responses (#91)

mikhirev:
Cache server failure responses for `cache-negative-ttl` seconds,
but not longer than 5 minutes.


folbricht (approved):

folbricht:
Thank you

Issue #116, opened 2020-12-28
Implement support for Oblivious DNS


folbricht:
As per https://tools.ietf.org/html/draft-pauly-dprive-oblivious-doh-03 and https://blog.cloudflare.com/oblivious-dns/

cbuijs:
Seems to work, didn't do extensive research/testing, and seems to be a tad slower then straight DoH to Cloudflare. Seems to retry a lot as well (not persistent?). But could be something external.

folbricht:
What sort of retries are you seeing? It has to reload the target key every few minutes as they expire but that should be it.

So far I haven't found a public (non-cloudflare) proxy to test with.

cbuijs:
Seems to work as expected, I recompiled everything from scratch and the retries are gone. I actually suspect something with my firewall wrong as I had some other issues as well (reboot fixed it, I think a state-table/memory problems was the cause, nothing related to RouteDNS).

mschirrmeister:
I was just trying the ODoH feature as well, but ran into a problem that I get no matter what the error `error="no key found for target \"odoh.cloudflare-dns.com\""`
I pulled the repo, merged #118 locally and build the docker image.

My config looks like this.
```
# DNS-over-HTTPS via Oblivious DNS

[resolvers.cloudflare-odoh-proxy]
address = "https://odoh1.surfdomeinen.nl/dns-query"
protocol = "odoh"
target = "https://odoh.cloudflare-dns.com/dns-query"

[listeners.local-udp]
address = "0.0.0.0:53"
protocol = "udp"
resolver = "cloudflare-odoh-proxy"

[listeners.local-tcp]
address = "0.0.0.0:53"
protocol = "tcp"
resolver = "cloudflare-odoh-proxy"
```

I have tried other proxies as well, but error is always the same. ODoH works in general if I test for example via https://github.com/cloudflare/odoh-client-go
Any hint what I am missing?

All my steps are the following.

```
git clone https://github.com/folbricht/routedns.git
cd routedns
git fetch origin pull/118/head:odoh
git checkout odoh
docker build -t routednsodoh .

docker run -it --name routedns \
    --rm \
    -p 5053:53/udp \
    -p 5053:53/tcp \
    -v /Users/marco/temp/odoh-client-simple.toml:/config.toml \
routednsodoh
```

Full error after sending a query.
```
[...]

PR #118, opened 2020-12-30
Support for Oblivious DNS client

folbricht:
Draft implementation of Oblivious DNS. Client-only for now if anyone wants to give it a try. The resolver config looks like this:

```toml
[resolvers.cloudflare-odoh-proxy]
# Address of the odoh proxy server. Also used to resolve the target's key
address = "https://1.1.1.1/dns-query"
protocol = "odoh"
# Address of the target. The hostname and path are passed to the proxy for forwarding
# of encrypted queries. No cert or bootstrap options for the target since the proxy
# connects to it on the client's behalf
target = "https://odoh.cloudflare-dns.com/dns-query"
```

Currently, I use Cloudflare as proxy and target which sort-of defeats the purpose of odns, but it's just for testing anyway. Don't know of any other proxies that support it yet.

Todo:
- [ ] DNSSEC validation of the target key response
- [x] Docs
- [x] Tests

Ref https://github.com/folbricht/routedns/issues/116


cbuijs:
The first post in this thread confuses me, the `target` is the ODoH proxy, and the `address` the "plain" DOH server (the DoH target using the ODoH proxy), right?

Might want to change `target` into `odoh-proxy` or something.

The cloudflare documentation on this confuses me as well, so it could be me :-).

folbricht:
hmm, I may have to work on the docs :) So the plain address is the "proxy", with "target" being the thing that will ultimately answer the encrypted query. The proxy has 2 jobs in RouteDNS:
1) Resolve the target's key, using a regular query. So it's not just a proxy, it's also a regular resolver (the DoH queries have different headers for proxy vs regular lookup)
2) Forward the encrypted query (encrypted by the client for the target) to the target and sending the response back the same way

Going to add a diagram when done with the implementation

cbuijs:
That is allright, the example works, but when people want to "disconnect" the proxy and the target (not being the same supplier for example), it might not be clear what to do, maybe point out with examples (we need more proxies though).

Seems that this is done by Cloudflare, Apple, Fastly, PCCW, equinix and SURF, none of them with the exception of Cloudflare offer anything yet as a proxy (I couldn't find any - yet).

Also not clear yet for me is that if any PROXY will actually work, as long as you publish the keys in DNS and TLS can be used as transport.

folbricht:
Yes, another public proxy would be great in the example, or a different target with Cloudflare as proxy.

It's actually much simpler, the *proxy* is always something that speaks DoH (the target too). The only difference is that it should recognize a different content header and additional query params to tell it to forward somewhere else. When it does forward, it does not work like a normal proxy (CONNECT) where the client is connected to the target and they talk. Instead, the proxy receives an encrypted query which is really just random-looking bytes from the proxy's point of view. It then just sends those bytes to the target and opens its own TLS connection to the target and again speaks DoH with the target sending those raw bytes (different content header again).. The client has no influence on the TLS link between proxy and target.

cbuijs:
Using "PROXY" in a "HTTPS" context is my muscle-brain memory kicking in for sure. :-)

alpominth:
Any plans to finish DNSSEC validation and make this branch ready?

folbricht:
Having a DNSSEC validator as part of RouteDNS would be awesome. Unfortunately it's also quite complex and I really don't have the time to work on it. Doing this would require at least a full week of work, possibly more. Can't say if I'll get to that any time soon.

cbuijs:
@folbright

Maybe check (if not done already), [GoResolver](https://pkg.go.dev/github.com/hazcod/goresolver)?

Note: Maybe we should continue this conversation at #123 ?

odohclient.go
```diff
@@ -0,0 +1,184 @@
...
+		return nil, time.Time{}, err
+	}
+
+	for _, answer := range response.Answer {
+		https, ok := answer.(*dns.HTTPS)
+		if !ok {
+			continue
+		}
+		for _, value := range https.Value {
+			if value.Key() != 32769 {
```
ignoramous, line 166: unfortunately, not all odoh servers publish https/svcb record. in fact, [RFC-9230](https://datatracker.ietf.org/doc/rfc9230/) has removed it from spec (from `draft-07` to be precise).

attempting to fetch odoh-config from both, `/.well-known/odohconfigs` and https/svcb records, is a safer bet. we do this for our android app, rethink ([code](https://github.com/celzero/firestack/blob/fb23a0fd4ff301701bf4d27d357c76473b9430f1/intra/doh/odoh.go#L142-L158)). cloudflare's reference implementation also does the same.

LeonardWalter:
hey, I updated this branch and implemented the config fetch from /.well-known/odohconfigs based on cloudflare's reference implementation on [my fork](https://github.com/LeonardWalter/routedns/tree/ODoH) 
I also added support for a proxy and ODoH server to routedns. 

It would be great if you could give some feedback.

folbricht:
@LeonardWalter Thanks for that. Would you like to drive dev of the odoh client in routedns? You could make a PR with the config changes against my PR and we merge it from there. The lack of a DNS validator is still an issue, perhaps best to call that out in the docs.

To give us a better starting-point, I rebased the PR and resolved conflicts. But unfortunately, `github.com/cloudflare/odoh-go` now has an indirect dependency on `git.schwanenlied.me/yawning/x448.git` which is down. So the branch is currently broken as I can't update `go.mod`. Not sure if that'll come back or if `github.com/cloudflare/odoh-go` is going to get updated to not depend on it.

folbricht:
As for feedback on your [fork](https://github.com/LeonardWalter/routedns/tree/ODoH). I only had a brief look at it so far. Pretty sure your implementation can be made a bit faster. It's re-doing a lot of the work for every request, like parsing the config. And you can use a `RWMutex` instead of the regular one to avoid requests having to wait.

LeonardWalter:
Yes, I can help with that :) I should have some time in the coming weeks.
I opened a PR with the first update to the ODoH client. I will make a new one for the server/proxy later on.

PR #141, opened 2021-03-23
Fix shuffle default for cache

### LINKED ISSUE #139: Cache has no default value for cache-answer-shuffle and breaks when not defined
Little bug, when no `cache-answer-shuffle` is provided on cache, routedns stops with the error message:

`Error: unsupported shuffle function ""`

Perhaps the value "static" should be allowed and selected as default?

P.S. Also seems to be some spelling-mistakes in the func-names (rando**n** instead of rando**m**).

folbricht:
Fixes #139


cbuijs:
Merci!

PR #144, opened 2021-03-26
Fix hang queuing a request

### LINKED ISSUE #143: Requests stuck when dialing a resolver takes too long
Finally I figured out the cause of my problem previously described in #140. It happens under heavy load when packets to one of upstream resolvers are dropped somewhere. It happens here:

https://github.com/folbricht/routedns/blob/2a699efccb74ff552f4016418f396d1055b2288a/pipeline.go#L48-L51

Queuing a request waits for all previous ones. If many requests are to be queued, their timeouts sum up.

mikhirev:
This fixes #143.


folbricht (approved):

folbricht:
Thank you.

PR #155, opened 2021-05-29
New fastest-tcp group to probe TCP connections

### LINKED ISSUE #149: Select "Fastest Server" feature
Tested some stuff with [Adguard DNSProxy](https://github.com/AdguardTeam/dnsproxy) and the `--fastest-addr` feature.

It actually works very well, and was wondering how much effort it is to build something similar for RouteDNS.

In short:
- The IP's of A or AAAA records coming in a response are tested with a Connect test and RTT of that is measured.
- Only the IP-Address with "fastest" RTT is returned to the DNS Client (and cached).

Some nice benefits: Connections seems to be more "snappy" on some sites and DNS responses are reduced in size (smaller packets on the network, less bandwidth).

A "con" is maybe that the first query, before caching takes potentially longer (because of the tests), but I didn't notice this from a "user" perspective. Also more "connects" flow out of my gateway due to the tests, but is negligible due to caching again.

Could be part of `response-collapse` maybe (make an option `fastest = true` or something)?

See also:
[Fastest addr + cache-min-ttl](https://github.com/AdguardTeam/dnsproxy#fastest-addr--cache-min-ttl).
[fastest_addr.go](https://github.com/AdguardTeam/dnsproxy/blob/63ee0fbc04e85f7b8492657c344950935fee1b30/fastip/fastest_addr.go)

folbricht:
Fixes #149

### DIFF
--- a/cmd/routedns/config.go
+++ b/cmd/routedns/config.go
@@ -111,6 +111,10 @@ type group struct {
 	Prefix6       uint8  // Prefix bits to identify IPv6 client
 	LimitResolver string `toml:"limit-resolver"` // Resolver to use when rate-limit exceeded
 
+	// Fastest-TCP probe options
+	Port    int
+	WaitAll bool `toml:"wait-all"` // Wait for all probes to return and respond with a sorted list. Generally slower
+
 	// Response Collapse options
 	NullRCode int `toml:"null-rcode"` // Response code if after collapsing, no answers are left
 }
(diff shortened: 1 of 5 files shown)


cbuijs:
Works as expected, maybe explain in documentation `success-ttl` has the same behaviour as `ttl-min` of cache? It kind of suggest it sets the TLL always.

folbricht:
It actually was always setting the TTL, but I now changed the option to `success-ttl-min` which does work like `ttl-min` of the TTL Modifier

Issue #158, opened 2021-06-02
Ability to add a time-frame when applicable to query/response/client blocklists


cbuijs:
It would be nice to have the ability when a blocklist group would be in effect time wise. Default would what it is now (always active), the following should be great to have:

- Starting Time
- Ending Time
- Which Days of the week (Sun, Mon, Tue etc), Default is all days.

When Ending Time is before Starting Time, it is assumed the "next day" as seen from the starting-time.

This give kind of a nice "Parental Control" option to block certain things during the day when kids are "awake", or Breakfast, Lunch and Dinner-time to force screens off (yes, I am a parent :-) ).

This could be also part of a new `group`, where you can select which resolver next based on time, or part of `router` perhaps... Router makes more sense as you could centralize the "schedules" in one place kind of.

Example:
```
routes = [
    {starttime = "18:00:00", endtime = "19:00:00", days = ["Mon", "Tue", "Wed"], resolver = "family-safe"},
    {starttime = "22:00:00", endtime = "07:00:00", resolver = "nxdomain"},
    {resolver = "anything-goes"},
]
```
In above rudimentary/abstract example, at certain time a "family-safe" resolver is used to block adult-themes or game-sites/domains (for example). During the night everything is blocked, and if those conditions are not met, the default "anything-goes" (no filtering) resolver is used.

PS: This is just a "nice to have".

folbricht:
Yes, good idea. I prefer the idea of a time-based router. That could be combined with any kind of other element, not just blocklist. Thinking if it's better in the existing router, or as a new element.

cbuijs:
Ah, I just updated the original post with some examples. :-)
Yeah, that makes sense, doesn't really matter for me, I kind of liked the "routers" way. That is all.

cbuijs:
BTW... There might be the need of a "flush-cache" feature on this. Because the TTL can go over the start-time for example and still serve answers maybe not wanted during the schedule.

One solution could be to use this always "before" cache (document).

Note: A flush-cache feature is cool anyway, also needed when lists are reloaded for example. I will make a separate post on/for that :-).

folbricht:
Time-based routing is now available on the `issue-158` branch if you want to give it a try. It supports routing by day of the week, as well as time of day. Here a simple example:
```toml
[routers.router1]
routes = [
[...]

PR #167, opened 2021-06-12
Support routing by time and weekday

folbricht:
Implements #158

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Features:
 - Support for plain DNS, UDP and TCP for incoming and outgoing requests
 - Connection reuse and pipelining queries for efficiency
 - Multiple failover and load-balancing algorithms, caching, in-line query/response modification and translation (full list [here](doc/configuration.md))
-- Routing of queries based on query type, query name, or client IP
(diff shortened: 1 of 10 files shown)


charlieporth1:
This will be a fantastic feature that I am looking forward to. I was going to implement this on my own servers using bash because I need this.

folbricht:
Just out of interest, do you have a specific use-case in mind for this? Is it also about routing to different blocklists based on the time of day or something more interesting? Just wondering if I should make an example for it and add to the docs.

charlieporth1:
> Just out of interest, do you have a specific use-case in mind for this? Is it also about routing to different blocklists based on the time of day or something more interesting? Just wondering if I should make an example for it and add to the docs.

I run a private DNS server on Google cloud and in order to save money with that I turn it off using another DNS server at home which they tend to use each other in failover events. When the Google cloud server shuts off I would like to route the traffic specially locally instead of the reduced topology I currently have that causes issues wth resolving

folbricht:
Thanks, interesting use-case. This PR is almost ready to go, just need a bit more documentation.

Issue #159, opened 2021-06-02
Ability to flush-cache


cbuijs:
There might be a need to flush-cache to make sure that any new entries come up in a blocklist reload might be still served due to TTL not expired yet.

Not sure if the cache(s) can be tracked back down the line/chain and flush them as a whole (rigourous). Or... maybe just for new blocklist entries loaded during a reload (which would be the safest way, and less impact, and most coolest).

See also #158 with similar effects.

folbricht:
This is actually quite hard to do using the current implementation as parts of the pipeline can't communicate with each other, apart from forwarding/receiving DNS queries. Don't have a good way to achieve this at the moment. Flushing is simple, but getting a signal in the right place isn't. I think it'll be best to address this with time-based routing as per #158 which is much cleaner and straightforward to implement.

cbuijs:
I solved this in my code to "listen" for specific queries by introducing a "command domain".

Lets say the "command domain" is "routedns.command", you can use dig to send "commands":

`dig @127.0.0.1 flush.routedns.command` - Flushes cache
`dig @127.0.0.1 reload.routedns.command `- Reload blocklists
etc...

Not sure how this would be possible in RouteDNS, but I guess a config option on the cache type, to define the "command domain" and then have pre-set values/triggers to do stuff. Same maybe for other types and relevant triggers/actions.

An empty "NOERROR" answer to show everything went well, of a REFUSED when not allowed, or a SERVFAIL when the command fails.

Not the most advanced or beautiful solution, but handy and practical. Only allow for localhost, as it is a nice gaping security whole when available to the world.

This is not a MUST HAVE feature, but come in handy to keep RouteDNS running instead of restarting which generates downtime. In a production environment could a lifesaver. Above is clunky, dirty and all of that, but seems to be an easy enabler.

folbricht:
That's quite a neat idea. Pushing it further, this could be turned into almost an API. Though at that point it may be best to simply add a REST API to the server to control components. A full API, allowing a dynamic re-configuration of elements would probably be a v2-thing though.

I should be able to add something simple like you described, flush the cache when a certain lookup is made

folbricht:
Draft implementation on `issue-159` if you want to give it a try. 

```toml
[groups.cloudflare-cached]
type = "cache"
resolvers = ["cloudflare-dot"]
cache-flush-query = "flush.cache."
```

You can send any query type.

PR #168, opened 2021-06-20
Support flushing a cache remotely

folbricht:
Implements #159


cbuijs:
Works as expected, maybe explain in documentation this has the same behaviour as `ttl-min` of cache?

charlieporth1 (approved):

Issue #170, opened 2021-07-15
UDP resolver should fallback to TCP on truncation.


mikhirev:
Hi!

This issue is about the same as #93, that was not properly fixed in fact.That "fix" just set EDNS buffer size to 4096 that does not completely prevent truncation but also introduces potential problems because of possible packet fragmentation.

- According to [DNS flag day 2020 recommendations][1], the minimum safe size is 1232 because it prevents fragmentation in most networks.
- Some servers, e.g. 1.1.1.1, ignore EDNS buffer size in request and truncate all large messages (bigger than 1232 in case of 1.1.1.1). You may check it with a `TXT` request for `test.knot-resolver.cz.`.
- [RFC7766][2] says:
  > - Recursive server (or forwarder) implementations MUST support TCP so that they do not prevent large responses from a TCP-capable server from reaching its TCP-capable clients.

UDP resolver should check the `TC` flag in answers and fallback to TCP if it is set. I also suggest to make EDNS buffer size configurable and set it to 1232 by default.

[1]: https://dnsflagday.net/2020/?#message-size-considerations
[2]: https://datatracker.ietf.org/doc/html/rfc7766#section-5

mikhirev:
This issue also affects DTLS resolver that should also have a fallback alternative. However [RFC8094] does not specify an exact fallback protocol:

> For Strict Privacy, the client MUST only send a new DNS request for the same resource record over an encrypted transport (e.g., DNS over TLS [RFC7858]). Clients using Opportunistic Privacy SHOULD try for the best case (an encrypted and authenticated transport) but MAY fall back to intermediate cases and eventually the worst case scenario (clear text) in order to obtain a response.

So it would be nice to provide a configuration option to set a custom resolver as a fallback for DTLS resolver.

[RFC8094]: https://datatracker.ietf.org/doc/html/rfc8094#section-5
[RFC7858]: https://datatracker.ietf.org/doc/html/rfc7858

folbricht:
There's a few things we should handle separately:

### Failover to TCP

[...]

PR #171, opened 2021-07-19
Support retry on truncation

folbricht:
Introduces `truncate-retry` group to support UDP truncation. Also improves handling of the TC bit when using UDP listeners as per #170

### DIFF
--- a/cache_test.go
+++ b/cache_test.go
@@ -173,3 +173,28 @@ func TestRoundRobinShuffle(t *testing.T) {
 	require.Equal(t, net.IP{0, 0, 0, 2}, a1.A)
 	require.Equal(t, net.IP{0, 0, 0, 1}, a2.A)
 }
+
+// Truncated responses should not be cached
+func TestCacheNoTruncated(t *testing.T) {
+	var ci ClientInfo
+	q := new(dns.Msg)
+	r := &TestResolver{
+		ResolveFunc: func(q *dns.Msg, ci ClientInfo) (*dns.Msg, error) {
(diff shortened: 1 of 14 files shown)


folbricht:
Nice, didn't notice that one. Using this now, thanks.

cache_test.go
```diff
@@ -173,3 +173,28 @@ func TestRoundRobinShuffle(t *testing.T) {
...
+			a.SetReply(q)
+			a.Truncated = true
+			return a, nil
+		},
+	}
+
+	c := NewCache("test-cache", r, CacheOptions{})
+
+	// Both queries should hit the upstream resolver
+	q.SetQuestion("test.com.", dns.TypeA)
```
charlieporth1, line 193: This domain `test.com` worries me unless you own the domain.
I would either query `www.google.com` or `github.com` for long term change and whitelist reasons. Those domains are much more "stable" as of today than `test.com` which I believe will exchange hands and possibly go `NXDOMAIN`, or `SERVFAIL` if push comes to shove. To add to that in my config I use pihole as the root lightweight blocker. If someone has `test.com` in there  base level DNS then that could add an issue as well depeneding on there config and setup if its blocked or not.

folbricht: Good point. While these queries don't actually hit any public resolvers, it's not good practice in general. I switched to using example.com instead which is intended to be used for testing.

charlieporth1: I just checked test.com is blocked on my config

charlieporth1 (approved):

Issue #152, opened 2021-05-27
new function request


iamsurfing:
https://github.com/iamsurfing/routedns/blob/master/lockduplicaterequest.go

iamsurfing:
Although there are still many bugs.

folbricht:
So what's this trying to do? It looks like it's slowing down/serializing requests for the same domain name, basically a name can only be queried sequentially. Is it meant to protect the upstream server? Or penalize the client? If it's about reducing load, one could use a short-term cache to reduce load on the upstream server.

Also, yes, there're still some bugs in the implementation :)

iamsurfing:
I use a lot of fastest group.

If network is not stable, or If resume from hibernating, routedns will make a DDoS attack to DNS server until receive response.

Most importantly, routedns will keep sending in-buffer-request even if cache store. Short-term cache can not solve this problem.

The target is locking duplicate request , then waiting cache store, then making duplicate request hit cache.

folbricht:
I'll try to improve on this to limit the performance-impact somewhat. The current implementation locks too much, and that would slow down non-duplicated queries as well.

folbricht:
There's now a draft implementation of this on the `issue-152` branch if you want to try it out. I haven't really done much testing on it yet.
The main difference between your implementation and this is that yours will hold requests and serialize them, while the `request-dedup` group on the branch will answer all in-flight requests with the first answer. So for example if 10 identical queries are sent, only one will actually be forwarded upstream and all 10 get the same answer.
Let me know what you think.

folbricht:
Oh, and this element might even be better used behind a cache, not in front. It should work either way though

iamsurfing:
I think count-to-10 is not best. I think timeout-after-5s is better.

downstream-->(this_component+cache)-->upstream

If get the first miss-cache request from downstream 
----a new lock and current timestamp will be stored in the map.
----request to upstream.
----After recieving a DNS answer
--------Request hit cache.
--------Clean this key-value from the map.
--------Set answer to map of cache.

If get duplicate request from downstream
----get the lock and current timestamp from the map.
----If the difference between stored timestamp and current timestamp less than 5s
--------It will get the lock from the map and lock it.
--------After receiving a DNS answer
------------Request hit the cache.
------------Clean this key-value from map.
------------Set answer to map of cache.
----Else [same as difference over than 5s]
--------It will unlock all locks, and answer SERVFAIL to all-downstream.
--------Clean this key-value from the map.

[...]

PR #181, opened 2021-08-22
New request-dedup group

folbricht:
Implements https://github.com/folbricht/routedns/issues/152


cbuijs:
Seems to be working as expected. I am missing some logging, I don't see it in the trace/debug-log as being hit/used? Maybe when an actual dedup is being issued, drop a line in the log as "dedupped"?

charlieporth1 (approved):

PR #183, opened 2021-09-10
Update configuration.md

charlieporth1:
Better documatation for `reset-after` clairifying usage.

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -417,7 +417,7 @@ Fail-Back groups are instantiated with `type = "fail-back"` in the groups sectio
 Options:
 
 - `resolvers` - An array of upstream resolvers or modifiers. The first in the array is the preferred resolver.
-- `reset-after` - Time in seconds to disable a failed resolver, default 60. Note this is not a timeout this disables the current failed resolver and moves on to the next one.
+- `reset-after` - Time in seconds to disable a failed resolver, default 60. Note: This is not a timeout argument. This disables the current failed resolver after x amount of time in seconds and moves on to the next one.
 - `servfail-error` - If `true`, a SERVFAIL response from an upstream resolver is considered a failure triggering a failover. This can happen when DNSSEC validation fails for example. Default `false`.
 
 #### Examples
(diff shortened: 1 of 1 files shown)


folbricht:
So the way this works is that if there's a failure in the preferred resolver this immediately switches to a failover resolver. Then, after `reset-after` seconds have passed, the group switches back to using the preferred resolver, not actually the next one in the list. It's the time it'll use alternative resolver before it resets back to the first. How about something like this? Not sure if that's more clear though:

`Time in seconds before switching from an alternative resolver back to the preferred resolver (first in the list) , default 60. Note: This is not a timeout argument. After a failure of the preferred resolver, this defines the amount of time to use alternative/failover 
resolvers before switching back to the preferred.`

charlieporth1:
> So the way this works is that if there's a failure in the preferred resolver this immediately switches to a failover resolver. Then, after `reset-after` seconds have passed, the group switches back to using the preferred resolver, not actually the next one in the list. It's the time it'll use alternative resolver before it resets back to the first. How about something like this? Not sure if that's more clear though:
> 
> `Time in seconds before switching from an alternative resolver back to the preferred resolver (first in the list) , default 60. Note: This is not a timeout argument. After a failure of the preferred resolver, this defines the amount of time to use alternative/failover resolvers before switching back to the preferred.`

I do like that more. Is there a limit of 2 to the number of resolvers? Or If you have more than 2 will It pass on to the next one like in my config 
```
# Order matters
[groups.ctp-dns-fail-back]
resolvers = [
        "ctp-dns-group",
        "ctp-dns-fail-rotate",
        "ctp-dns_nginx-back-up-group-fail-back",
        "ctp-dns-failover-backup",
]
reset-after = 128
type = "fail-back"
servfail-error = true

[groups.ctp-dns-fail-rotate]
resolvers = [
        "ctp-dns_group-fastest-home-external",
        "ctp-dns_group-fastest-aws-external",
        "ctp-dns_group-fail-rotate-home",
        "ctp-dns_group-fail-rotate-aws",
        "ctp-dns_group-fail-back-home",
        "ctp-dns_group-fail-back-aws",
]
type = "fail-rotate"
servfail-error = true
```

folbricht:
Yes, you can have more than one, and those are specified in order of preference basically. Imagine you had a list of resolvers like so

```
[r0, r1, r2, r3,...]
```

In this case, r0 is the "preferred" resolver. If r0 fails, the group switches to r1, if that one fails it moves on to r2 and so on. After the reset interval expires, r0 becomes the active resolver again. So it always starts with r0, and then moves further one step on failure, and starting from the beginning again after the given time.

folbricht (approved):
Thank you

PR #184, opened 2021-09-16
Added versioning for updating and debugging reasons

charlieporth1:
This will be my first time coding in go so tell me how I did and change what you need to at the EOD it's your code your project. Added versioning for updating and debugging reasons


cmd/routedns/main.go
```diff
@@ -68,6 +72,10 @@ func start(opt options, args []string) error {
 	if opt.logLevel > 6 {
 		return fmt.Errorf("invalid log level: %d", opt.logLevel)
 	}
+	if opt.version {
+		printVersion()
+		os.Exit(0)
+	}
```
folbricht, line 78: Since MinArgs is set to 0 now, I think it'd be best to add a check for the number of args here. Something like:

```golang
if len(args) < 1 {
  return errors.New("not enough arguments")
}
```

cmd/routedns/main.go
```diff
@@ -716,3 +724,13 @@ func newIPBlocklistDB(l list, locationDB string, rules []string) (rdns.IPBlockli
 		return nil, fmt.Errorf("unsupported format '%s'", l.Format)
 	}
 }
+
+func printVersion() {
+	var versionString interface{} = rdns.BuildVersion
+	var buildNumber interface{} = rdns.BuildNumber
+	var buildTime interface{} = rdns.BuildTime
```
folbricht, line 731: Any reason to declare it as an `interface{}` here? Wouldn't it be possible to just do
```golang
  fmt.Println("Build:",  rdns.BuildVersion)
  ...
```

version.go
```diff
@@ -0,0 +1,8 @@
+package rdns
+
+var (
+	BuildVersion string = "3.0.0"
```
folbricht, line 4: Do you have a way to generate this file during the build? Just wondering why it's at `3.0.0` already.

cmd/routedns/main.go
charlieporth1: We can change that

version.go
charlieporth1: I was going to add that in a different PR because of using GitHub actions. It is currently 3.0.0 because the version was not set and it was not 1.0.0.

cmd/routedns/main.go
```diff
@@ -68,6 +72,15 @@ func start(opt options, args []string) error {
 	if opt.logLevel > 6 {
 		return fmt.Errorf("invalid log level: %d", opt.logLevel)
 	}
+	if opt.version {
+		printVersion()
+		os.Exit(0)
+	} else {
```
folbricht, line 78: You don't really need the "else" path here.

version.go
folbricht: I can tag a proper release when I'm back next week. Shall we say `v0.1.0` since it's the first one?

folbricht (approved):

version.go
charlieporth1: Sure

folbricht:
It's still a draft, but once it's ready and merged I'll tag master with `v0.1.0`

PR #195, opened 2021-12-12
Default ports feature

charlieporth1:
Default ports if ports are not specified


cmd/routedns/example-config/well-known-wo-ports.toml
```diff
@@ -0,0 +1,149 @@
+# This file shows a selection of well-known DNS resolvers and supported protocols to
+# help build RoutedDNS configs. It isn't intended to be used directly but can for testing.
+# Ports are set in config because they are set to the default protocols port
```
folbricht, line 3: Can you clarify this sentence? Should it say "not set in config"?

cmd/routedns/main.go
```diff
@@ -192,9 +197,15 @@ func start(opt options, args []string) error {
 
 		switch l.Protocol {
 		case "tcp":
-			listeners = append(listeners, rdns.NewDNSListener(id, l.Address, "tcp", opt, resolver))
+			if portIsSet == false {
+				listenAddress = listenAddress + rdns.PlainDNSPort
+			}
+			listeners = append(listeners, rdns.NewDNSListener(id, listenAddress, "tcp", opt, resolver))
```
folbricht, line 203: You can simplify all this quite a bit. Just have a function that adds the port if it's not already on it. Something like 
```go
func addressWithDefault(addr, defaultPort string) string
```
That would let you change all instances to something like
```go
listenAddress = addressWithDefault(listenAddress, rdns.PlainDNSPort)
```
You could even do that inline if the lines don't get too long (probably better to keep it separate).

Another thing, when you're looking for `:` in an address, please use https://pkg.go.dev/net#SplitHostPort and https://pkg.go.dev/net#JoinHostPort. Then you can also remove the colon from the constants below.

static_defaults.go
```diff
@@ -0,0 +1,10 @@
+package rdns
+
+var (
+	DoQPort      string = ":8853"
```
folbricht, line 4: No need for the colon if your function uses https://pkg.go.dev/net#JoinHostPort from the standard library

cmd/routedns/main.go
charlieporth1: Done

static_default_ports.go
```diff
@@ -0,0 +1,24 @@
...
+	DoTPort      string = "853"
+	DTLSPort     string = DoTPort
+	DoHPort      string = "443"
+	PlainDNSPort        = "53"
+)
+
+func AddressWithDefault(addr, defaultPort string) string {
+	_, port, _ := net.SplitHostPort(addr)
+	var isPortEmpty bool = len(port) > 0
+	if isPortEmpty == false {
```
folbricht, line 19: The idiomatic way to do this in Go is to:
```go
if port == "" {
```
and you don't need the `else` path

cmd/routedns/resolver.go
```diff
@@ -55,6 +64,9 @@ func instantiateResolver(id string, r resolver, resolvers map[string]rdns.Resolv
 			return err
 		}
 	case "doh":
+		r.Address = rdns.AddressWithDefault(r.Address, rdns.DoHPort)
+		r.BootstrapAddr = rdns.AddressWithDefault(r.BootstrapAddr, rdns.DoHPort)
+
```
folbricht, line 69: Not sure this is the right way to do it. The addresses can be URLs I think, and BootstrapAddr isn't (and shouldn't) be set for everything like this. I'll come back to this later

charlieporth1: I thought about that after I published it. Maybe we create an issue for it and push it for now knowning it can be a bug

folbricht: I wonder if it's sufficient for the function to return a blank string if the address is empty. Like

```go
if addr == "" {
  return addr
}
```

That would preserve blank values which are the default

static_default_ports.go
charlieporth1: Agreed will do

cmd/routedns/resolver.go
charlieporth1: I like that idea!

charlieporth1: To improve on this do you think it would be better if it was just the host part of SplitHostPort because of the `:853` etc

static_default_ports.go
```diff
@@ -0,0 +1,36 @@
...
+var (
+	DoQPort      string = "8853"
+	DohQuicPort  string = "1443"
+	DoTPort      string = "853"
+	DTLSPort     string = DoTPort
+	DoHPort      string = "443"
+	PlainDNSPort        = "53"
+)
+
+func AddressWithDefault(addr, defaultPort string) string {
```
folbricht, line 19: Sorry to keep bringing up this function. Since you're now dealing with different address formats, you can consider using https://pkg.go.dev/net/url#Parse to do the heavy lifting, it's going to be a lot more reliable. Consider doing it like this:

```go
if addr == "" {
  return addr
}
u, err := url.Parse(addr)
if err != nil {
  return addr
}
if u.Scheme == "http" || u.Scheme == "https" {
  return addr // do we want to set a port on those? Maybe not
}
if u.Port() == "" {
  u.Host := net.JoinHostPort(u.Host, defaultPort)
}

return u.String() // re-assemble the address, now with port

Comment (charlieporth1) [reply static_default_ports.go:19]: I think we do want to set a port on `http` and `https` tho, because there are alt ports for both such as `8080` or `8443` which both require that scheme to be processed correctly

Comment (folbricht) [reply static_default_ports.go:19]: Can certainly do that, it shouldn't hurt anything. I revised the code snippet from above to do that. I think it'll work, should add a test for it.
```go
if addr == "" {
  return addr
}
u, err := url.Parse(addr)
if err != nil {
  return addr
}
if u.Port() == "" {
  u.Host = net.JoinHostPort(u.Host, defaultPort)
}
if u.Scheme == "" { // no url, just host+port
  return u.Host
}
return u.String() // re-assemble the address, now with port
```

charlieporth1: I run into this error - 
```
zsh: exit 1     ./routedns example-config/well-known-wo-ports.toml
```
do to - 
```
[resolvers.cloudflare-doh-1-0-0-1-get]
address = "https://1.0.0.1/dns-query{?dns}" --- here
protocol = "doh"
doh = { method = "GET" }
```
I have no idea on a good way to fix it

folbricht: Did you use my code above? I can't repro this here. You can commit what you have and I can take a look

charlieporth1: I pretty much did this is what I used 
```
if addr == "" {
		return addr
	}
	u, _ := url.Parse(addr)
	if u.Port() == "" {
		u.Host = net.JoinHostPort(u.Host, defaultPort)
	}
	if u.Scheme == "" { // no url, just host+port
		return u.Host
	}
	return u.String() // re-assemble the address, now with port
``` 
I removed the error part after I figured out the `{}` was causing an error when parsing the url

folbricht: Turns out this had way more edge-cases than I expected. So I added a test and simplified the function a bit. It's https://github.com/folbricht/routedns/pull/202 which targets your branch. Feel free to take it as is onto your branch, or build on top of it. I may also have missed a few test cases, should be easy to add those now

folbricht (approved):
Thank you

PR #202, opened 2021-12-28
Test for default-ports feature and simplify the function

folbricht:


cmd/routedns/main.go
```diff
@@ -229,9 +229,9 @@ func start(opt options, args []string) error {
 			listeners = append(listeners, ln)
 		case "doh":
 			if l.Transport != "quic" {
-				l.Address = rdns.AddressWithDefaultForHttp(l.Address, rdns.DoHPort)
+				l.Address = rdns.AddressWithDefault(l.Address, rdns.DoHPort)
 			} else if l.Transport == "quic" {
-				l.Address = rdns.AddressWithDefaultForHttp(l.Address, rdns.DohQuicPort)
+				l.Address = rdns.AddressWithDefault(l.Address, rdns.DohQuicPort)
 			}
 			tlsConfig, err := rdns.TLSServerConfig(l.CA, l.ServerCrt, l.ServerKey, l.MutualTLS)
```
charlieporth1, line 236: Is this supposed be a change from `AddressWithDefaultForHttp` to `AddressWithDefault` or is that out of date?

cmd/routedns/resolver.go
```diff
@@ -61,7 +62,7 @@ func instantiateResolver(id string, r resolver, resolvers map[string]rdns.Resolv
 			return err
 		}
 	case "doh":
-		r.Address = rdns.AddressWithDefaultForHttp(r.Address, rdns.DoHPort)
+		r.Address = rdns.AddressWithDefault(r.Address, rdns.DoHPort)
```
charlieporth1, line 66: Same here

cmd/routedns/main.go
folbricht: It's intentional. I changed the function to support both formats, http and regular endpoints. That way you can use the same for both cases. Also added a few tests if you want to confirm this does handle both.

charlieporth1: Ok

Issue #208, opened 2022-02-02
Ability to force DNS client to use TCP (set TC bit)


cbuijs:
See also #205 and https://github.com/folbricht/routedns/issues/205#issuecomment-1027650690.

To mitigate some common DDOS attacks, which mostly rely on UDP and Spoofing, a feature that kicks in based on particular load, that the DNS clients can be told to switch to TCP. This can be done by the DNS Server (RouteDNS) answering with the `TC` (Truncate) bit set. It annulls spoofing and takes the punch out of the attacks relying on it (almost all DNS based attacks).

I think having a `static-responder` (like the `drop` one), where you can set this is enough to mitigate and use it together with the `rate-limiter` for example.

Would look like this as example:
```
# Rate-Limiter
[groups.rrl]
type = "rate-limiter"
resolvers = ["internet-resolver"]
limit-resolver = "set-tc"
requests = 100
window = 120
prefix4 = 24
prefix6 = 56

# Respond with TC bit set
[groups.set-tc]
type = "truncate"
```

Result will be that a query of a DNS client will be responded to with an "Try again using TCP". Two effects, traffic will go down a bit as only legitimate clients (unspoofed ones) can connect, and when TCP is used, more mitigation/standard features (congestion control, queing, etc) of the TCP stack will kick in as standard and can be used as well.

See also: https://help.fortinet.com/fddos/4-3-0/FortiDDoS/Understanding_FortiDDoS_DNS_Attack_Prevention.htm

cbuijs:
I have thrown something together, see PR #209

cbuijs:
Tested it, and works:

```
dig @<censored> -p 9053 www.paypal.com
;; Truncated, retrying in TCP mode.

; <<>> DiG 9.16.23 <<>> @<censored> -p 9053 www.paypal.com
; (2 servers found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 49979
;; flags: qr rd ra ad; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;www.paypal.com.			IN	A

;; ANSWER SECTION:
www.paypal.com.		2400	IN	A	151.101.65.21

;; Query time: 10 msec
;; SERVER: <censored>#9053(<censored>)
;; WHEN: Wed Feb 02 11:36:12 CET 2022
;; MSG SIZE  rcvd: 62
```

[...]

PR #209, opened 2022-02-02
Add type "truncate" group

cbuijs:
See #208 

Might need to review as this is my copy-paste programming skills with low Golang affinity :-).


truncate.go
```diff
@@ -0,0 +1,31 @@
...
+// NewTruncateResolver returns a new instance of a TruncateResolver resolver.
+func NewTruncateResolver(id string) *TruncateResolver {
+        return &TruncateResolver{id}
+}
+
+// Resolve a DNS query by returning a fixed response.
+func (r *TruncateResolver) Resolve(q *dns.Msg, ci ClientInfo) (*dns.Msg, error) {
+        answer := new(dns.Msg)
+        answer.SetReply(q)
+        answer.Truncated = true // This is the bit!
```
folbricht, line 25: Do you want logging here?

folbricht:
My first thought was that this could be added to the static-responder, since it's quite similar. Thoughts?

truncate.go
cbuijs: Good one. Yes please! :-)

cbuijs:
See comment here: https://github.com/folbricht/routedns/issues/208#issuecomment-1028693775

PR #211, opened 2022-02-04
Add truncate option in static-responder to set TC bit

cbuijs:


cbuijs:
See also #208

cmd/routedns/example-config/truncate.toml
```diff
@@ -0,0 +1,28 @@
...
+protocol = "tcp"
+resolver = "doh-google"
+
+#### THE MAGIC HAPPENS HERE ####
+[groups.truncate]
+type = "static-responder"
+rcode = 0 # NOERROR
+truncate = true
+
+#### Bootsrap ####
```
folbricht, line 19:
```suggestion
#### Bootstrap ####
```

doc/configuration.md
```diff
@@ -925,6 +926,14 @@ routes = [
 ]
 ```
 
+Retunr emtp answer with TC (Truncate) bit set so DNS client is instructed to retry the query using TCP instead of UDP.
```
folbricht, line 929:
```suggestion
Return an emtpy answer with TC (Truncate) bit set so the DNS client is instructed to retry the query using TCP instead of UDP.
```

static.go
```diff
@@ -51,28 +53,37 @@ func NewStaticResolver(id string, opt StaticResolverOptions) (*StaticResolver, e
...
+		// Update the name of every answer record to match that of the query
+		answer.Answer = make([]dns.RR, 0, len(r.answer))
+		for _, rr := range r.answer {
+			r := dns.Copy(rr)
+			r.Header().Name = qName(q)
+			answer.Answer = append(answer.Answer, r)
+		}
+		answer.Ns = r.ns
+		answer.Extra = r.extra
+		answer.Rcode = r.rcode
```
folbricht, line 84: I think it might be best to preserve the existing logic and just adding the truncate flag to it. So just adding

```go
answer.Truncated = r.truncate
```
should do the trick. That way if there are other things like NS records etc, they aren't ignored when truncate is set. That makes it more flexible. As for logging, you could use `logger(r.id, q, ci).WithField("truncated", r.truncated).Debug("responding")` perhaps. Then you don't need the if...else condition

cbuijs: I like it, let me see how I can edit this stuff... :-)

cbuijs: Changed it, including the logging and removing the if condition.

cbuijs:
@folbricht Could you review the changes? I tested it here and all works as expected.

folbricht (approved):
Looking good, thank you

PR #232, opened 2022-04-10
Update dependencies

folbricht:
- Update all library dependencies to the latest. There was a breaking change in `quic-go` which required minor code and terminology updates
- Also set the min Go version to 1.17


charlieporth1 (approved):

charlieporth1:
Good fix that I didn't catch. Good job!

PR #247, opened 2022-07-01
Prefetch feature

charlieporth1:


cache-prefetch.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+
+type CachePrefetchMetrics struct {
+	// Cache hit count.
+	domainEntries map[string]CachePrefetchEntry
+}
+type CachePrefetchEntry struct {
+	// request hit count.
+	hit 			  expvar.Int
```
folbricht, line 27: Probably best to stick to the base types, and make the whole entry thread-safe.

cache-prefetch.go
```diff
@@ -0,0 +1,183 @@
...
+
+	// 0 for no prefetching job
+	// 1 for prefetching job is active
+	// 2 for stopped by error
+	prefetchingStatus expvar.Int
```
folbricht, line 32: Instead of documenting what each value means, you can use a dedicated type.

```go
type PrefetchState int

const (
  PrefetchStateNone = iota
  PrefetchStateActive
  PrefetchStateOther
)
```

cache-prefetch.go
```diff
@@ -0,0 +1,183 @@
...
+		return nil, errors.New("no question in query")
+	}
+	// While multiple questions in one DNS message is part of the standard,
+	// it's not actually supported by servers. If we do get one of those,
+	// just pass it through and bypass caching.
+	if len(q.Question) > 1 {
+		return r.resolver.Resolve(q, ci)
+	}
+
+	go r.requestAddPrefetchJob(q, ci)
```
folbricht, line 92: It's not a good idea to start a goroutine per query, that can easily overload the server. This can instead be done in a single goroutine or keep it synchronous.

cache-prefetch.go
```diff
@@ -0,0 +1,183 @@
...
+	// a copy since other elements might modify the response, like the replacer.
+	return a, nil
+}
+
+func (r *CachePrefetch) String() string {
+	return r.id
+}
+func (r *CachePrefetch) startCachePrefetch(q *dns.Msg, ci ClientInfo) {
+	var maxNumberOfErrorsBeforeDiscardingPrefetchJob = int64(5)
+	var qname = q.Question[0].Name
```
folbricht, line 110: Access to the map is not actually thread-safe this way. You'd need a mutex for this.

cache-prefetch.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+	}
+}
+func (r *CachePrefetch) requestAddPrefetchJob(q *dns.Msg, ci ClientInfo) {
+	if len(q.Question) < 1 {
+		return
+	}
+	var qname = q.Question[0].Name
+	r.metrics.domainEntries[qname] = CachePrefetchEntry{}
+	var domainEntry = r.metrics.domainEntries[qname]
```
folbricht, line 145: Not sure I understand what this is trying to do, first it flushes the record with a blank one, then it reads that empty struct into `domainEntry`?

cache-prefetch.go
```diff
@@ -0,0 +1,183 @@
...
+
+func (r *CachePrefetch) String() string {
+	return r.id
+}
+func (r *CachePrefetch) startCachePrefetch(q *dns.Msg, ci ClientInfo) {
+	var maxNumberOfErrorsBeforeDiscardingPrefetchJob = int64(5)
+	var qname = q.Question[0].Name
+
+	for {
+		var domainEntry = r.metrics.domainEntries[qname]
```
folbricht, line 113: I bet we need to include the query type in that key, not just the qname (otherwise it's going to treat queries for MX and A records the same). And possibly even the values from the opt records, take a look at how the cache generates a key.

folbricht (commented):
I didn't go through all of it yet since it likely needs a couple of major changes.

Rather than starting a goroutine per query, it's better to synchronously update the access-counter. And have a single goroutine just iterate over the list to find the ones that should be fetched, similar to how the GC routine in the cache clears out old entries.

cache-prefetch.go
charlieporth1: I was trying to initiatize it

cache-prefetch.go
charlieporth1: Let me know if my new version is any good

cache-prefetch.go
charlieporth1: Its been done in a single goroutine

cache-prefetch.go
charlieporth1: I like this better

cache-prefetch.go
charlieporth1: Are you saying `int32` instead of `expvar.Int`??

cache-prefetch.go
```diff
@@ -0,0 +1,221 @@
...
+	PrefetchStateOther  = 2
+)
+
+type CachePrefetchEntry struct {
+	// request hit count.
+	hit expvar.Int
+
+	prefetchState PrefetchState
+	// store the time to live
+	msg *dns.Msg
```
charlieporth1, line 43: Unfortunately, this does make the cache double up. Is there a better way to cache a DNS question? I tried just storing the DNS question which would result in a `SERVFAIL` is there more I need to store or less?

charlieporth1:
@folbricht I think it should be all good for you to review again. I would like you to review it in draft before I add it to full PR because I would like it to run on my servers as a test as well

cache-prefetch-metrics.go
```diff
@@ -0,0 +1,135 @@
...
+
+// 0 for no prefetching job
+// 1 for prefetching job is active
+// 2 for stopped by error
+type PrefetchState int
+
+const (
+	PrefetchStateNone   = iota
+	PrefetchStateActive = 1
+	PrefetchStateOther  = 2
```
folbricht, line 24:
```suggestion
	PrefetchStateActive
	PrefetchStateOther
```

cache-prefetch-metrics.go
```diff
@@ -0,0 +1,135 @@
...
+	if item != nil {
+		return item.prefetchState
+	}
+	// Add new item to the top of the linked list
+	if len(c.items) > c.maxItems {
+		return PrefetchStateOther
+	}
+
+	item = NewCachePrefetchEntry(key)
+	c.items[key] = item
```
folbricht, line 81: This looks like it's not safe to be called from multiple goroutines.

cache-prefetch-metrics.go
```diff
@@ -0,0 +1,135 @@
...
+	return item.prefetchState
+}
+func (c *CachePrefetchMetrics) addHit(query *dns.Msg) {
+	key := c.getDomainKey(query)
+	item := c.items[key]
+	item.hit ++
+	if item.hit >= c.hitMin {
+		item.prefetchState = PrefetchStateActive
```
folbricht, line 89: This too isn't threadsafe.

cache-prefetch-metrics.go
```diff
@@ -0,0 +1,135 @@
...
+	item.errorCount = 0
+	c.items[key] = item
+}
+// Loads a cache item
+func (c *CachePrefetchMetrics) touch(key CachePrefetchKey) *CachePrefetchEntry {
+	item := c.items[key]
+	if item == nil {
+		return nil
+	}
+	return item
```
folbricht, line 114: All of this is basically equivalent to
```go
return c.items[key]
```

cache-prefetch-metrics.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (c *CachePrefetchMetrics) getDomainKey(q *dns.Msg) CachePrefetchKey {
+	if (q == nil) || len(q.Question) < 1 {
+		return CachePrefetchKey{}
+	}
+	qname := qName(q)
+	qtype := qType(q)
+	key := CachePrefetchKey{qname, qtype}
+	return key
+}
```
folbricht, line 135: Dobn't think you need these two functions. Neither of them use anything from `r` or `c` so they don't need to be receivers and you can just call a single function

cache-prefetch.go
```diff
@@ -0,0 +1,149 @@
...
+				r.mu.Unlock()
+				Log.WithFields(logrus.Fields{"qname": qname, "qtype": qtype}).Trace("query reset error count")
+			}
+			Log.WithFields(logrus.Fields{"qname": qname, "qtype": qtype}).Debug("query prefetched")
+		}
+	}
+}
+
+func (r *CachePrefetch) requestAddPrefetchJob(q *dns.Msg) {
+	if (q == nil) || len(q.Question) < 1 {
```
folbricht, line 142: No need to check the length of the question slice again I think, it's already done in the Resolve function

cache-prefetch.go
```diff
@@ -0,0 +1,149 @@
...
+
+func (r *CachePrefetch) String() string {
+	return r.id
+}
+func (r *CachePrefetch) startCachePrefetchJobs() {
+	Log.WithFields(logrus.Fields{"id": r.id}).Trace("starting prefetching job")
+	for {
+		time.Sleep(r.CacheTTLPollingCheckInterval)
+		domainEntriesLength := len(r.metrics.items)
+		if domainEntriesLength > 0 {
```
folbricht, line 102: No need for the `if` here, the `for` loop will not iterate if it's empty

cache-prefetch.go
```diff
@@ -0,0 +1,149 @@
...
+	if item.prefetchState == PrefetchStateActive { // only prefetch if status is 1
+		qname := qName(item.msg)
+		qtype := qType(item.msg)
+
+		Log.WithFields(logrus.Fields{"qname": qname, "qtype": qtype}).Trace("prefetch request started")
+		var ci ClientInfo
+		a, err := r.resolver.Resolve(item.msg.Copy(), ci)
+
+		if err != nil || a == nil {
+			Log.WithFields(logrus.Fields{"err": err}).Trace("prefetch error")
```
folbricht, line 124: You'd normally use `Log.WithError(err).Trace(...)`

cache-prefetch-metrics.go
folbricht: I see the mutex you added to the caller. Not sure this will be safe with the prefetch loop iterating over the items.

One thing you could try is add the mutex to `CachePrefetchMetrics` and lock/unlock as needed within its methods. That shifts responsibility for locking into CachePrefetchMetrics. But even then you need to be careful with `range r.metrics.items` without a lock.

folbricht (commented):
I haven't had time to go through it fully yet. But noticed at least one race condition. Also there seem to be unrelated changes in this PR. A bunch of files are now empty, GH workflows deleted, and `.idea` files added. Was any of that intentional?

charlieporth1:
> I haven't had time to go through it fully yet. But noticed at least one race condition. Also there seem to be unrelated changes in this PR. A bunch of files are now empty, GH workflows deleted, and `.idea` files added. Was any of that intentional?

Not at all. that was due to a rebase issue

cbuijs:
Tried to fiddle around with this, but I do not see anything prefetched. Is there some kind of logging I can check for?

charlieporth1:
> Tried to fiddle around with this, but I do not see anything prefetched. Is there some kind of logging I can check for?

Because my dumb a** forgot to add a `dns.msg` to the query prefetch request so they were all `nil`. :/
This has since been fixed

cbuijs:
> > Tried to fiddle around with this, but I do not see anything prefetched. Is there some kind of logging I can check for?
> 
> Because my dumb a** forgot to add a `dns.msg` to the query prefetch request so they were all `nil`. :/ This has since been fixed

No worries :-). Will give it a try again later.

folbricht:
Have you considered moving this feature into the existing cache? Since there's only a couple config flags this should be quite straightforward. I can give this try if you like.

Also, the PR now contains a whole bunch of permission changes which makes this quite hard to read.

charlieporth1:
> Have you considered moving this feature into the existing cache? Since there's only a couple config flags this should be quite straightforward. I can give this try if you like.
> 
> Also, the PR now contains a whole bunch of permission changes which makes this quite hard to read.

I was thinking that originally.
Sorry that on the permissions changes WSL and Windows caused that because I was building on WSL and committing on Windows which for some really odd reason was changing the permissions

folbricht:
If you don't mind, I'd like to take your branch and try to merge the prefetch into the existing cache. It might make it easier to use, definitely going to use your changes and just shuffle them around a bit and see if it works.

charlieporth1:
Np!

thb007:
configuration There is no introduction to this feature .https://github.com/folbricht/routedns/blob/master/doc/configuration.md#Static-responder

folbricht:
There's an similar feature being tested on https://github.com/folbricht/routedns/tree/prefetch-3 at the moment (it's still draft so no docs there either).

PR #254, opened 2022-08-13
Add support for ARM in Dockerfile

mrmenndev:
Right now inside the Dockerfile `GOARCH` is hardcoded to use `amd64`.
With this pull request I created the script called `dockerbuild.sh` wich determine if the CPU is `amd64`, `arm` or `arm64`, 
so the docker version of routedns can be used on ARM based plattform (for example the Raspberry PI).


cmd/routedns/dockerbuild.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+set -e
+
+arch=$(uname -m)
+
+if [ "$arch" = "x86_64" ]; then
```
charlieporth1, line 6: Hate to be picky but this should probably be a `case` - `esac` function

mrmenndev: Allright done

folbricht:
This assumes that the build host is the same architecture as the host where the output is to be run. Wouldn't it be better to define an ARG in the dockerfile and simply pass in the desired value with `--build-arg arch=...`? See https://docs.docker.com/engine/reference/builder/#arg

mrmenndev:
I thought about defining `ARG GOARCH` in the Dockerfile, 
so if you want to build for a different type of ARCH it will use this value, otherwise it should fallback to the build host arch because docker don't enforce defining `--build-args`
What do you think ?

folbricht:
I'd flip this around, basically like:
- Define the GOARCH (and maybe add GOOS too) as build args, using amd64 and linux as defaults
- Leave the build command in the docker file, using the args
- Wrap the docker build command in a shell script like the one in this PR, so the build system can decide what system to build for and set the args. It's going to be a lot cleaner than trying to auto-detect inside the build container

Would that work? It seems more flexible since it'd handle the OS too.

mrmenndev:
Seems to be the simplest solution.
Would the build script called like ` ./dockerbuild.sh "linux" "arm64" ` if you want to build for a different arch ?

folbricht:
That works. And in the script you can default to "current" arch while still allowing people to call the docker build directly on windows for example.

mrmenndev:
Should the script still try to autodetect the arch if you don't define any value ?
For example `./dockerbuild.sh` "arm64" would build for arm64, while `./dockerbuild.sh` would use the arch from host.

folbricht:
That's sort of it's only purpose at this point. But we should ask if it's even still needed? It's not normally something required to wire up a build system. Actually, doesn't go default to "current" if GOARCH and GOOS are empty? So if that's the case, all you'd have to do is remove the default value from the new ARG in the dockerfile.

mrmenndev:
I guess we should keep the values empy and let GO decide. If the container need to be build for a diffent arch or os the user can always define `GOARCH` or `GOOS`.

folbricht (approved):
Thank you

PR #279, opened 2023-01-26
Implement cache-prefetch

folbricht:
Implementation of a basic cache-prefetch feature similar to how Bind does it https://kb.isc.org/docs/aa-01122

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -58,6 +58,14 @@ type CacheOptions struct {
 
 	// Query name that will trigger a cache flush. Disabled if empty.
 	FlushQuery string
+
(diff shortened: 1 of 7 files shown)


alpominth:
Two important options included in branch https://github.com/folbricht/routedns/compare/master...prefetch-feature-2 are missing:

cache-ttl-polling-check-interval
record-query-hits-min

https://github.com/folbricht/routedns/commit/15df284fe9224e9170b38300f60f225f12a6ada0#diff-69854e85865fdfa4a40d619bcfe214fedcba863aadf7c1a55062ab7f827991f6R19

PR #281, opened 2023-02-02
Bump quic-go to 0.32.0

tobyxdd:


charlieporth1:
> 

Please provide a description so we can test accordingly

adminlistener.go
```diff
@@ -9,8 +9,8 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/lucas-clemente/quic-go"
-	"github.com/lucas-clemente/quic-go/http3"
+	"github.com/quic-go/quic-go"
```
charlieporth1, line 12: Why was this library chosen over the previous one

charlieporth1 (approved):

adminlistener.go
tobyxdd: It's the same library, they just changed its package name.

tobyxdd: See https://github.com/quic-go/quic-go/issues/3611 for more information

Issue #286, opened 2023-02-23
addition option to configure query tiemout


brotaxt:
Hey there! 

Is there any way to raise the query timeout? 

From time to time I get some "timed out" errors with routedns. If we could get an option to raise the timeout value, that could probably help to mitigate those errors.

```
routedns[20539]: time="2023-02-23T03:00:11+01:00" level=error msg="failed to resolve" addr="[::1]:8053" client="::1" error="query for 'fw-download.ubnt.com.' timed out" id=local-doh-behind-proxy_ipv6 path=/dns-query protocol=doh qname=fw-download.ubnt.com. qtype=AAAA
routedns[20539]: time="2023-02-23T03:04:25+01:00" level=error msg="failed to resolve" addr="[::1]:8053" client="::1" error="query for 'vg01.met.vgwort.de.' timed out" id=local-doh-behind-proxy_ipv6 path=/dns-query protocol=doh qname=vg01.met.vgwort.de. qtype=AAAA
routedns[20539]: time="2023-02-23T03:04:25+01:00" level=error msg="failed to resolve" addr="[::1]:8053" client="::1" error="query for 'scdn.cxense.com.' timed out" id=local-doh-behind-proxy_ipv6 path=/dns-query protocol=doh qname=scdn.cxense.com. qtype=AAAA
```

folbricht:
That's a fair request. It's currently fixed to a value given in the RFC but should be configurable. It's set here https://github.com/folbricht/routedns/blob/master/pipeline.go#L14 if you want to try out a few values.

I'm currently wondering if it should be a global value, or configurable on a resolver-level. Being able to provide a setting per resolver is probably better.

brotaxt:
I would also prefer a setting per resolver :)

folbricht:
https://github.com/folbricht/routedns/pull/295 adds a `query-timeout` option to all resolvers that lets you specify the number in seconds. The default is also 2s now if this isn't set. Would you be able to try it out and see if that addresses the issues?

cbuijs:
Works here without any issues, but using different config:
```
[resolvers.nextdns-doh3]
protocol = "doh"
Aaddress = "https://doh3.dns.nextdns.io/xxxxxx/dns-query"
transport = "quic"
bootstrap-address = "45.90.28.0"
query-timeout = 5
```
Using the anycast address in the bootstrap and "POST" method.

cbuijs:
[...]

PR #287, opened 2023-02-25
close UDP socket after quic new stream fails

mattkeenan:
- we need to close the UDP socket after a new quic session fails before we attempt to retry with a new connection, this is because we created the UDP socket not quic-go, so we have to clean up our own mess as quic-go won't clean it up for us
  - remove pool for doq connections as this is now fixed in quic-go, c.f. https://github.com/quic-go/quic-go/issues/3718
  - update quic-go version in go.mod
  - WARNING: updated quic-go REQUIRES go 1.19 or newer to compile


mattkeenan:
It would be nice to get some further testing on this; I've tested locally, but it is a relatively simple set up, one desktop & one laptop, doq from adguard, cache, and udp 53 listener / resolver for local network.

Also, it's only been tested on Linux here, not Windows nor MacOS.

I can write some test functions as well if required, but this will need separate tests on Linux, Windows, MacOS (which I might need contributions on, or I can just set up nop tests on those platforms).

mattkeenan:
And lastly, this is just a first contribution to this repo, so feedback would be appreciated.

charlieporth1:
@mattkeenan Appreciate the contributions. I will get on this tmr

dohclient.go
```diff
@@ -275,66 +273,113 @@ func dohQuicTransport(endpoint string, opt DoHClientOptions) (http.RoundTripper,
...
+	stream, err := s.Connection.OpenStreamSync(ctx)
+	if err != nil {
+		// Try to open a new connection, but clean up our mess before we do so
+		_ = s.Connection.CloseWithError(DOQNoError, "")
+		// and then we need to close the connection / socket ourselves as we own
+		// the UDP socket not quic-go
+		// c.f. https://github.com/quic-go/quic-go/issues/1457
+		_ = s.udpConn.Close()
+		s.udpConn = nil
+		logrus.Infoln("temporary fail when trying to open stream, attempting new connection")
```
folbricht, line 333: Should this really be Info-level since we're handling this case (and it's quite common). You could also attach the error like so:
```go
Log.WithError(err).Debug("temporary fail when trying to open stream, attempting new connection")
```
This will also use the built-in logger (also logrus).

dohclient.go
```diff
@@ -275,66 +273,113 @@ func dohQuicTransport(endpoint string, opt DoHClientOptions) (http.RoundTripper,
...
+
+		var connection quic.Connection
+		connection, s.udpConn, err = quicDial(s.hostname, s.rAddr, s.lAddr, s.tlsConfig, s.config)
```
folbricht, line 336: If there's an error here as `s.udpConn` is set to nil in the struct then it'll panic on the next attempt when calling `_ = s.udpConn.Close()`. I think you need to do the same as with `connection` here.

dohclient.go
```diff
@@ -275,66 +273,113 @@ func dohQuicTransport(endpoint string, opt DoHClientOptions) (http.RoundTripper,
...
+	defer s.mu.Unlock()
+	stream, err := s.Connection.OpenStreamSync(ctx)
+	if err != nil {
+		// Try to open a new connection, but clean up our mess before we do so
+		_ = s.Connection.CloseWithError(DOQNoError, "")
+		// and then we need to close the connection / socket ourselves as we own
+		// the UDP socket not quic-go
+		// c.f. https://github.com/quic-go/quic-go/issues/1457
+		_ = s.udpConn.Close()
+		s.udpConn = nil
```
folbricht, line 332: This could be a problem when the next attempt fails as well since there's going to be a nil-ptr panic above. We could either make sure it's never nil, or add a condition

go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/oschwald/maxminddb-golang v1.10.0
 	github.com/pion/dtls/v2 v2.2.4
 	github.com/pkg/errors v0.9.1
-	github.com/quic-go/quic-go v0.32.0
+	github.com/quic-go/quic-go v0.33.0
```
folbricht, line 14: Since it requires Go 1.19 now, can you bump the version at the top of the file?

doqclient.go
```diff
@@ -144,7 +143,7 @@ func (d *DoQClient) Resolve(q *dns.Msg, ci ClientInfo) (*dns.Msg, error) {
 
 	// Write the query into the stream and close is. Only one stream per query/response
 	_ = stream.SetWriteDeadline(time.Now().Add(time.Second))
-	if _, err = stream.Write(p); err != nil {
+	if _, err = stream.Write(b); err != nil {
```
folbricht, line 146: Whoops, that must be my bug. Thanks for the fix!

doqclient.go
```diff
@@ -224,9 +223,16 @@ func (s *doqConnection) getStream() (quic.Stream, error) {
...
 	if netErr, ok := err.(net.Error); ok && (netErr.Timeout() || netErr.Temporary()) {
-		// Try to open a new connection
+		// Try to open a new connection, but clean up our mess before we do so
 		_ = s.connection.CloseWithError(DOQNoError, "")
-		s.connection, err = quicDial(s.hostname, s.endpoint, s.lAddr, s.tlsConfig, s.config, s.pool)
+		// and then we need to close the connection / socket ourselves as we own
+		// the UDP socket not quic-go
+		// c.f. https://github.com/quic-go/quic-go/issues/1457
+		_ = s.udpConn.Close()
+		s.udpConn = nil
```
folbricht, line 232: Same comment as above, just make sure we don't run into a nil-ptr panic here ever.

folbricht:
This is great, thanks for working on this. Left some minor comments.

dohclient.go
```diff
@@ -275,66 +273,113 @@ func dohQuicTransport(endpoint string, opt DoHClientOptions) (http.RoundTripper,
...
 }
 
-// Wrapper for http3.RoundTripper due to https://github.com/quic-go/quic-go/issues/765
-// This wrapper will transparently re-open expired connections. Should be removed once the issue
-// has been fixed upstream.
-type http3ReliableRoundTripper struct {
-	*http3.RoundTripper
-	pool *udpConnPool
+func (s *quicConnection) HandshakeComplete() context.Context {
+	return s.expiredContext
```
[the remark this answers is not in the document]
mattkeenan: Sorry, this looks like my mistake; dohclient used to use *Early connections (from much earlier versions) and not any more, the HandshakeComplete is only used for Early style connections; I'll go through and fix it up. From my reading as we're not using Early style connections we have to use the default 5s timeout; for my tastes that is a bit long for a DNS lookup, but I'm happy with it if you are. Just let me know your preference and I can move it all to default connections, or move it to *Early connections (where we could implement non default timeouts if you want). If I'm misunderstanding something here feel free to point me in the right direction.

dohclient.go
mattkeenan: You're right, I'll put a guard around it.

dohclient.go
mattkeenan: Yeah, I'll put a guard around it like above.

go.mod
mattkeenan: Will do

doqclient.go
mattkeenan: np

doqclient.go
mattkeenan: Yeah, will do

mattkeenan:
@folbricht I suspect you'll squash my commits when you merge; do you prefer individual commits for each fix? Or is a single rollup commit ok?

folbricht:
You can just put it all in one, no problem. It'll get squashed during merge.

dohclient.go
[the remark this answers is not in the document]
folbricht: Either works for me. It's been a while since I looked at the differences between them. You probably have a better idea what's best at this point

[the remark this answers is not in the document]
mattkeenan: Do we have a default timeout across all transport types? or do we default to whatever timeout the transport defaults to? Sorry for all the questions, my knowledge of routedns overall is a bit shallow so far.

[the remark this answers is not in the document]
folbricht: We currently have one global query timeout, that'll soon change to become a per-resolver timeout (see https://github.com/folbricht/routedns/issues/286). As for connection timeouts, we don't really have anything that applies to all transports.

mattkeenan:
Hi,
I made the changes as requested. There are a few minor points;

1. I did some minor refactoring
  1.1.  Removed the doqConnection struct as all the fields in this struct were just copies from either DoQClient or quicConnection
  1.2. Created a quicRestart() function to restart a quicConnection
    1.2.1. Previously this was done inline in 3 places with similar but different logic (DRY)
2. I decided to use quic.DialEarly as this allows us to use 0-RTT for quicker queries
  2.1. first packet sends the query, response will be in first reply (if server supports it)
  2.2. I removed the HanshakeComplete handling as the default should work fine for us
    2.2.1. We may need to adjust this in future if we come across servers that have borked 0-RTT rejection handling

And lastly, I found a bug in doqclient with setting the query DNS msg header Id to 0; this fails if you copy the query to multiple upstream resolvers (like fastest groups do) as second and following queries will have their query Id set to 0 by the first query. I'll create a separate branch / PR for this that depends on this branch.

mattkeenan:
Sorry, I forgot to bump the go version to 1.19, fixed now.

mattkeenan:
@folbricht bump!

dohclient.go
```diff
@@ -275,66 +273,139 @@ func dohQuicTransport(endpoint string, opt DoHClientOptions) (http.RoundTripper,
...
 
-// UDP connection pool. Also a workaround for for the http3.RoundTripper. When using a custom
-// dialer that open its own UDP connections, http3.RoundTripper doesn't close them when the
-// remote terminates a connection, or when calling Close(). So we have to keep track of the
-// connections and close them all before calling Close() on the http3.RoundTripper.
-type udpConnPool struct {
-	conns []*net.UDPConn
-	mu    sync.Mutex
+func (s *quicConnection) NextConnection() quic.Connection {
+	return nil
```
folbricht, line 348: Is this safe? I mean if it's called and used by the caller. Or should it return `s`?

dohclient.go
```diff
@@ -275,66 +273,139 @@ func dohQuicTransport(endpoint string, opt DoHClientOptions) (http.RoundTripper,
...
 }
 
-func (p *udpConnPool) closeAll() {
-	if p == nil {
-		return
+func quicDial(hostname, rAddr string, lAddr net.IP, tlsConfig *tls.Config, config *quic.Config) (quic.EarlyConnection, *net.UDPConn, error) {
+	udpAddr, err := net.ResolveUDPAddr("udp", rAddr)
+	if err != nil {
+		Log.WithError(err).Debug("couldn't resolve remote addr (" + rAddr + ") for UDP quic client")
```
folbricht, line 394: No need to change it now, but there's a Debugf() that may come in handy, same in the log below

folbricht: Or you could use
```go
Log.WithError(err).WithField("remote", rAddr).Debug("couldn't resolve address")
```

folbricht:
Looking good, thank you. Left some comments but will merge this in for now so if there's anything you want to change you can do that in the next one with the QueryID fix or so.

dohclient.go
mattkeenan: Yeah, you're right. I'll send through a clean up patch later. Thanks.

dohclient.go
mattkeenan: No, I wouldn't have thought so, the purpose of this function is to suggest a fallback connection downgrade (similar to downgrading from TLS v1.3 to TLS v1.2). Since we handle the one time fallback ourselves (with the quicReconnect) I thought it best to have no fallback (even falling back to ourselves, which would probably put us into a loop), hence returning `nil`. If I'm wrong then let me know and I'll send a clean up patch.

folbricht: You're probably right. That function isn't documented upstream, and luckily also not used from what I can see. There's a test that does this
```go
newConn := conn.NextConnection()
str, err := newConn.OpenUniStream()
```
It'd panic if `newConn` was nil, but again, it's not used so seems fine

PR #289, opened 2023-03-03
update README for DoQ info and golang ver requirement bump

mattkeenan:
A quick README update;

- Add updated info about DoQ; RFC & 0-RTT
- routedns now requires golang 1.19+


folbricht (approved):

folbricht:
Thank you

Issue #153, opened 2021-05-28
RouteDNS exits when using an URL "source" that generates an error


cbuijs:
When using a `source` pointing to an URL, and the download fails (for example 404: Does not exist), RouteDNS exits.

Would it be possible to exit more gracefully, or have an option to ignore the error and retry again later? And stay running?

iamsurfing:
I think you can download to local disk by a script, and load from local disk.

folbricht:
It's possible to do that, just need to figure out a way to do it elegantly. This is already working if the file disappears, routedns will keep running and use the old content until it's able to refresh again. I deliberately wanted to make it fail if the blocklist isn't there on startup, but that could probably be made configurable.

cbuijs:
use the `cache-dir` statement, so in most caches it reverts to the local version when not expired. But sometimes when I start from scratch, and maybe made a typo in the URL, or it is unavailable due to whatever reason (404 for example, or timeout).

It might be handy indeed to have an option to "continue" anyway. The reload/refresh should still be done, but maybe at a shorter/definable interval in that case? In that case maybe a retry option makes sense as well.

cbuijs:
Ping!

If we can make this configurable, to make it continue when an error happens on download? Perhaps with a fail-over using multiple sources?

folbricht:
Yup, I can get that in the next few days. Just been quite short on time.

cbuijs:
No hurries!

cbuijs:
Ping! :-)

cbuijs:
Pong? :-)

charlieporth1:
@cbuijs I think I can work on it. Give me a couple of days but that will free up @folbricht's time

folbricht:
I whipped up a quick draft implementation in https://github.com/folbricht/routedns/pull/259 This only addresses the `domain` loader, not the other 2 yet. There's a new `allow-fail-on-startup` option on the list config.

```toml
blocklist-source = [
   {format = "domain", source = "https://raw.githubusercontent.com/cbuijs/accomplist/master/deugniets/routedns.blocklist.domain.list1", allow-fail-on-startup = true},
```

It's completely un-tested. @charlieporth1 Would you like to take care of the other loaders? You can see from that PR how I planned on dealing with it. Any suggestions are welcome. Feel free to push to the same branch.

charlieporth1:
I would like to

cbuijs:
Works.

folbricht:
[...]

PR #294, opened 2023-03-25
New 'allow-failure' flag for blocklist loaders

folbricht:
Implements #153 

Adds a new `allow-failure` bool flag to avoid failure on startup


charlieporth1 (approved):

cbuijs:
Tested, it works. Together with the other modifications on blocklists, the crashes are gone.

PR #295, opened 2023-03-25
New 'query-timeout' option for resolvers

folbricht:
Implements #286 

New flag that can be used to set a query timeout.
```toml
[resolvers.cloudflare-dot]
address = "1.1.1.1:853"
protocol = "dot"
query-timeout = 4 # seconds
```


charlieporth1 (approved):

cbuijs:
Seems to work nicely. I always have retries/timeouts with Quad9, and with `query-timeout = 5`, that is gone now.

Issue #299, opened 2023-04-14
Make "cache" persistent


cbuijs:
It would be cool, especially on larger environments, to be able to make `cache` persistent between reloads of RouteDNS and don't lose it. This to avoid a huge uptick or burst in load on environments with high number of queries when reloading/restarting.

A common way doing so is to flush the cache every X amount of time to disk/file, and this file is loaded when RouteDNS (re)starts. TTLs stay ticking, so on startup the purge of the cache(file) needs to happen, based on the records original time-stamp and TTL left, to avoid stale/old records to be served.

Config example (last 3 lines):
```
[groups.persitent-cache]
type = "cache"
resolvers = ["google-doh"]
cache-size = 8192
cache-negative-ttl = 60 # Seconds
gc-period = 30 # Seconds
cache-harden-below-nxdomain = true
cache-flush-query = "flush.cache."
cache-prefetch-trigger = 10 # Seconds
cache-prefetch-eligible = 30 # Seconds
persistent = true
persistent-save-period = 300 # Every 5 minutes
cache-file = "/opt/routedns/persistent.cache"
```

**Note:** Maybe even an option to "preload" the cache with our own records. This would enable generated/synthesized/spoofed answers for redirection, blocking, etc...

cbuijs:
I agree. Not really care how it is done, the more flexibility the better :-). For me is not loosing cache between reloads is the heavy requirement. Sharing/Merging the cache among different instances of RouteDNS using Redis for example is quite cool too. It again reduces overhead and duplication.

folbricht:
There's now a draft implementation of persistence on the `issue-229` branch. You can configure it like so
```toml
[groups.cloudflare-cached]
type = "cache"
resolvers = ["cloudflare-dot"]
backend = {type = "memory", filename = "/var/tmp/cache.json"}
```
It'll read from the file on startup and overwrite it on shutdown. The old configs should be supported (could you confirm?) Let me know how that works.

cbuijs:
Will try coming days when I find some time. Does it keep track of TTL on their original timestamp?

Example, if a record is cached with TTL of 60 seconds and the cache get reloaded more than 60 seconds later it is gone, right?

folbricht:
It's an exact dump of the state of the cache, so items can definitely "age" while routedns is shut down. These will be loaded again on startup, but get dropped shortly after.

cbuijs:
Seems to work as expected.

Two things so far:

[...]

PR #300, opened 2023-04-23
Implement memory backend with persistence

folbricht:
Adds a new "backend" concept to caches, with currently only one type of backend (memory) supported

```toml
[groups.cloudflare-cached]
type = "cache"
backend = {type = "memory", size = 1000, filename = "/var/tmp/cache.json"}
```

Implements #299


charlieporth1:
Do you need someone to test this?

folbricht:
Yes please. It should be complete at this point, but I did only basic testing. This should be the first supported "backend" for caches.

charlieporth1:
What kill signal should routedns get to write on close

charlieporth1:
> What kill signal should routedns get to write on close

Dumb question sigint sighup

folbricht:
That's fixed, it should now save on SIGTERM as well

folbricht:
Not shared currently. It'd be possible to point to the same file and restart etc, but even that can fail since they won't be sync'ed and can step on each other.

There's another open issue to implement a redis backend which would provide this sync. The "memory" backend is just the first step.

folbricht:
Not at this point. I didn't really want to use a signal just for one element. There could be other uses for SIGHUP. But you can configure an short interval for saving the cache for testing

folbricht:
Redis-support will come after this, but it'll have to wait a few weeks since I won't be able to work on it until the end of May.

folbricht:
Sorry about that. It was caused by another change which conflicted. Pushed a fix

folbricht:
That should probably be `"c:/users/..."` or `"c:\\users\\..."`

charlieporth1:
@alpominth I did a rebuild and no issues found I think it was due to to much traffic

folbricht:
Thank you for the extensive testing.

PR #307, opened 2023-05-02
Sub domain replace feature

charlieporth1:
No doc or unit test done yet will provide when time is abundant


cmd/routedns/example-config/subdomain-replace.toml
```diff
@@ -0,0 +1,29 @@
+# Basic DNS proxy. Translate all plain DNS queries received on port 53
```
folbricht, line 1: old comment

subdomain-replace.go
```diff
@@ -0,0 +1,115 @@
...
+	} else {
+		return "", false
+	}
+}
+func (r subDomainReplaceExpressions) apply(qname string) string {
+	for _, e := range r {
+		s, result := e.substitute(qname)
+		if result {
+			return s
+		} else {
```
folbricht, line 47: Blank `else` here

subdomain-replace.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// NewReplace returns a new instance of a Replace resolver.
+func NewSubDomainReplace(id string, resolver Resolver, list ...SubDomainReplaceOperation) (*SubDomainReplace, error) {
+	var exp subDomainReplaceExpressions
+	for _, o := range list {
+		if strings.Contains(o.From, "{}") && strings.Contains(o.To, "{}") {
+			exp = append(exp, subDomainReplaceExp{o.From, o.To})
+		} else {
+				return nil, errors.New("{} not found")
```
folbricht, line 65: A few formatting issues, please run it through `gofmt`

subdomain-replace.go
```diff
@@ -0,0 +1,115 @@
...
+	to   string
+}
+
+type subDomainReplaceExpressions []subDomainReplaceExp
+func (r subDomainReplaceExp)  substitute (qname string) (string, bool) {
+	 fromDomain := r.from
+	toDomain := r.to
+	// from {}.ctptech.dev to {}.charlesp.tech
+	// from test.ctptech.dev to test.charlesp.tech
+	fromDomain = strings.Replace(fromDomain, "{}", "", 1)
```
folbricht, line 32: Since this only works for the last part of a domain anyway, you wouldn't even need the `{}` placeholder at all. It could simply be `from=some.domain.net to=example.com`

folbricht (commented):
Does this actually do anything that the existing "replace" doesn't? https://github.com/folbricht/routedns/blob/master/doc/configuration.md#Replace looks more flexible as it's possible to replace any part of the domain using regexes

charlieporth1:
> Does this actually do anything that the existing "replace" doesn't? https://github.com/folbricht/routedns/blob/master/doc/configuration.md#Replace looks more flexible as it's possible to replace any part of the domain using regexes

No, it doesn't but it is simpler which is why I created it

subdomain-replace.go
```diff
@@ -0,0 +1,113 @@
...
+
+// NewReplace returns a new instance of a Replace resolver.
+func NewSubDomainReplace(id string, resolver Resolver, list ...SubDomainReplaceOperation) (*SubDomainReplace, error) {
+	var exp subDomainReplaceExpressions
+	for _, o := range list {
+		if strings.Contains(o.From, "{}") && strings.Contains(o.To, "{}") {
+			exp = append(exp, subDomainReplaceExp{o.From, o.To})
+		} else {
+			return nil, errors.New("{} not found")
+		}
```
folbricht, line 64: Do you still need this since `{}` isn't necessary in the config anymore?

subdomain-replace.go
```diff
@@ -0,0 +1,113 @@
...
+	from string
+	to   string
+}
+
+type subDomainReplaceExpressions []subDomainReplaceExp
+
+func (r subDomainReplaceExp) substitute(qname string) (string, bool) {
+	fromDomain := r.from
+	toDomain := r.to
+	// from {}.ctptech.dev to {}.charlesp.tech
```
folbricht, line 31:
```suggestion
	// from .ctptech.dev to .charlesp.tech
```

subdomain-replace.go
```diff
@@ -0,0 +1,113 @@
...
+	// from test.ctptech.dev to test.charlesp.tech
+	if strings.HasSuffix(qname, fromDomain) {
+		str := strings.Replace(qname, fromDomain, toDomain, 1)
+		Log.WithField("qname", qname).Debug("matches: modifying query")
+		return str, true
+	} else {
+		return "", false
+	}
```
folbricht, line 39:
```suggestion
	} 
	return "", false
```

subdomain-replace.go
```diff
@@ -0,0 +1,113 @@
...
+
+type subDomainReplaceExpressions []subDomainReplaceExp
+
+func (r subDomainReplaceExp) substitute(qname string) (string, bool) {
+	fromDomain := r.from
+	toDomain := r.to
+	// from {}.ctptech.dev to {}.charlesp.tech
+	// from test.ctptech.dev to test.charlesp.tech
+	if strings.HasSuffix(qname, fromDomain) {
+		str := strings.Replace(qname, fromDomain, toDomain, 1)
```
folbricht, line 34: I think there's a small issue with just doing a simple replace since the "fromDomain" string can show up more than once in the domain. Let's say you have `fromDomain = .local.` and the domain name being looked up is `example.local.something.local.` then it'd replace the first occurrance. It might be best to simply strip the `fromDomain` with `strings.TrimSuffix()`, then add the `toDomain` value to the result with `+` like so:
```go
sub := strings.TrimSuffix(qname, fromDomain)
str := sub+toDomain
return str, true
```

folbricht (commented):
One thing I'm not to fond of is the same. This isn't replacing "sub"-domains, but actually it's replacing domains. Should this be called `domain-replace` perhaps?

Also, could you add a section to the documentation? This is effectively a more efficient and simple replacing group, faster than the regexp replacer.

charlieporth1:
For sure

folbricht:
If you want to use `{}`, it'd make sense to change the behavior of the replacing logic and allow replacing in any position, not just at the front of the name. Otherwise using `{}` is confusing since it looks like one can place it anywhere, but it's only really supported in one place, and only a single instance.

Issue #278, opened 2023-01-23
[Feature Request] Redis DNS cache


alpominth:
A Redis backed DNS cache would be very good to have implemented in routedns.

The cache would be persistent and would persists across reboots.

These projects have this feature already implemented:

[Unbound](https://github.com/NLnetLabs/unbound)
[Blocky](https://github.com/0xERR0R/blocky)
[Overture](https://github.com/shawn1m/overture)
[mosdns-cn](https://github.com/IrineSistiana/mosdns-cn)
[GODNS](https://github.com/kenshinx/godns)
[redins](https://github.com/arvancloud/redins)

This package could serve as a basis for this implementation: https://github.com/go-redis/redis

alpominth:
@folbricht

Do you consider my suggestion? I found RouteDNS the most advanced DNS forwarder written in Go, and this idea to implement caching of DNS queries with Redis is overkill to me. I don't think it's difficult to implement, you could just follow the steps in another project here: https://github.com/shawn1m/overture/commit/5f9e7b12b8239f18defd8caec35e09e8f2fa3139

(I am not a GO programmer, if I were an I would have already implemented this idea even if it was only for personal usage only).

folbricht:
It can certainly be implemented, probably not too difficult. I'm a bit short on time but can take a quick look at it.

alpominth:
Thanks, no pressure, no problems, I have been using Unbound for persistent caching with Redis but it requires bothering modifications on Redis.

Changing the subject, a very nice and useful feature would be prefetching, it's already implemented on separate branches, if you could merge into the master would be a good idea: https://github.com/folbricht/routedns/compare/master...prefetch-feature-2

folbricht:
Weird, prefetch was supposed to be merged already, just re-opened a PR https://github.com/folbricht/routedns/pull/279, missing docs still

alpominth:
Very good, but I noticed that two important options of branch https://github.com/folbricht/routedns/compare/master...prefetch-feature-2 are missing:

cache-ttl-polling-check-interval = 60
record-query-hits-min = 1

==> https://github.com/folbricht/routedns/commit/15df284fe9224e9170b38300f60f225f12a6ada0#diff-69854e85865fdfa4a40d619bcfe214fedcba863aadf7c1a55062ab7f827991f6R19

It would be very good for who visits sites periodically and want to keep the cache up to date. Together with Redis cache would be overkill.

folbricht:
[...]

PR #315, opened 2023-05-27
Implement Redis backend for cache

folbricht:
Support for storing caches in Redis to allow for multiple instances sharing a common cache.

```toml
[groups.cloudflare-cached]
type = "cache"
backend = {type = "redis", redis-address = "127.0.0.1:6379" }
```

Implements #278

### DIFF
--- a/cache-redis.go
+++ b/cache-redis.go
@@ -3,7 +3,7 @@ package rdns
 import (
 	"context"
 	"encoding/json"
-	"fmt"
+	"errors"
 	"strings"
 	"time"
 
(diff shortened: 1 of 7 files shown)


alpominth:
I tested it, I performed some queries with dig app, no problems.

I will make extensive tests, but I will wait to some additional features be implemented.

I also tested with a Redis unix socket, again, no problems.

The next feature would be the ability of setting the password for Redis: https://github.com/redis/redis/blob/unstable/redis.conf#L1035-L1044 - https://github.com/redis/redis/blob/unstable/redis.conf#L863-L957

/\ If an additional effort could be put for implementing authenticating an user together with password, it would be nice.

PS: @folbricht My intention is not to pressure you, implement the features when you have a spare time if you wish.

alpominth:
@folbricht 

One question, at this time, does RouteDNS with this PR is able to automatically able to delete expired cached items from Redis?

folbricht:
Password auth should already work (not tested it but it's implemented). You can even set a different RedisDB if you want. See the doc changes:

```
- `redis-network` - The network type, either `tcp` or `unix`. Defaults to `tcp`.
- `redis-address` - Address of redis database, host:port
- `redis-password` - Redis password
- `redis-db` - Redis database to be selected
```

As for expiring items from the cache, that is also implemented. Cached records are dropped (by Redis) when their TTL runs out. I can quickly add a `redis-username` too, that's trivial

alpominth:
@folbricht 

Wonderful.

The only features that I think it would be useful are ability to set cooldown between Redis connections attempt and ability to specify the number of connections attempts.

It could be:

redis-connection-attempts: <number of attempts> # 0 is infinite
redis-connection-cooldown: <number of seconds>

Let's suppose that RouteDNS starts with SystemD after Redis and tries to connect immediately, in slow systems maybe Redis cannot set up the listeners before RouteDNS calls for it. This is reasonable.

folbricht:
I'm not familiar with Redis' restart behavior specifically, but I could wire up any of these options in the Redis client: https://pkg.go.dev/github.com/redis/go-redis/v9#Options

Perhaps these are relevant:

```
// Maximum number of retries before giving up.
// Default is 3 retries; -1 (not 0) disables retries.
MaxRetries int

// Minimum backoff between each retry.
// Default is 8 milliseconds; -1 disables backoff.
MinRetryBackoff time.Duration

// Maximum backoff between each retry.
// Default is 512 milliseconds; -1 disables backoff.
MaxRetryBackoff time.Duration
```

It doesn't really say what `MaxRetries = 0` means though.

alpominth:
I tried to select the redis-db 1 here but I got this:

```
$ /dev/shm/routedns /dev/shm/permcache.toml 
Error: toml: line 19 (last key "groups.dnscrypt-cached.backend.redis-db"): incompatible types: TOML value has type string; destination has type integer
```

folbricht:
This should be a number in the config, not a string. Without double-quotes it should work `redis-db = 1`

alpominth:
Got.

It worked now.

alpominth:
I will made a bulk and stress test in this branch and I will post the results here.

alpominth:
Here the tests, I tested with AF_INET6 and AF_UNIX:

**Over Unix Socket:**

Config 1: https://paste.ee/p/P7LIa
Config 2: https://paste.ee/p/AuRie
Results: https://paste.ee/p/PyMUs

**Via Redis standard port ([::1]:6379):**

Config 1: https://paste.ee/p/FuRFa
Config 2: https://paste.ee/p/AuRie
Results: https://paste.ee/p/uKI2i

==================================

Result: No segmentation fault in any RouteDNS instance. Read/Write errors in the benchmark app with 1024 concurrent DNS requests in parallel but I believe it should be like this because of overhead and demand of processing (my CPU is "domestic").

Compiled and tested in a AMD Ryzen 5 1400 4 cores 8 threads, 2133Mhz DDR4 RAM, Debian GNU/Linux amd64, go version go1.19.6 linux/amd64.

PS: I believe this branch is ready for merging into the master one, even under a heavy loading, no significant problem was detected.

alpominth:
I have been using this branch for 4 days, no problems so far.

The features that remain to be implemented are writing of key prefixes in Redis DB, and maybe, ability for specifying the number of retries in Redis connection and connection cooldown interval between connection attempts.

Also, an interesting feature that it would be interesting to implement is support for Redis Sentinel:

https://redis.io/docs/reference/sentinel-clients/

https://redis.io/docs/management/sentinel/

/\ I think it would be easy to implement, as you could just see and commit in RouteDNS code based on a ready implementation (Blocky): https://github.com/search?q=repo%3A0xERR0R%2Fblocky+sentinel&type=code

folbricht:
The following options have been added, which should include all of the features, except for Sentinel. Only did basic testing of the prefix, and no testing of the retry flags.
```
redis-key-prefix
redis-max-retries
redis-min-retry-backoff
redis-max-retry-backoff
```

I think Sentinel is something we could implement in another feature request/PR if needed.

alpominth:
@folbricht 

Personally, I have no use for Redis Sentinel, just having Redis support is enough for me. But you never know if someone needs this feature now.

I will compile and test for the upcoming days, it's a pitty that I cannot install on Windows as Redis shouldn't run, because I manage a dozen of PCs with Windows 10.

alpominth:
More tests:

Over Unix Socket:

Config 1: https://paste.ee/p/mzsEc
Config 2: https://paste.ee/p/KPn6j
Results: https://paste.ee/p/aUmdT

Via Redis standard port ([::1]:6379):

Config 1: https://paste.ee/p/uy1lw
Config 2: https://paste.ee/p/KPn6j
Results: https://paste.ee/p/pwD60

==================================

No problems, the test using 1024 concurrent requests had some failures due to the enormous amounts of DNS requests and overdhead, but this is expected.

I don't know why, but the serving of DNS queries of the cache is slower if the response is NXDOMAIN, it happens both with RAM and Redis cache backend.

@folbricht

I believe this branch is ready to merge into **master**.

folbricht:
It's not immediately obvious why caching of NXDOMAIN would be noticably slower, unless you use `cache-harden-below-nxdomain` which I don't see in the config. Maybe `cache-rcode-max-ttl` being set to 0 in config 2 could do that. Would need to test it.

alpominth:
I modified the config 2 and added cache-harden-below-nxdomain: https://paste.ee/p/pni2u

Results: https://paste.ee/p/ETTjq

Again, the same result.

But this doesn't matter, NXDOMAIN are rarely served by a DNS resolver.

charlieporth1 (approved):

bjester:
Very nice! Seems like the performance of this was well tested. Although, I'm curious what the performance difference might be compared with the existing memory caching?

alpominth:
@bjester 

I made some tests with RAM backend: https://github.com/folbricht/routedns/pull/300#issuecomment-1544357059

You may compare with https://github.com/folbricht/routedns/pull/315#issuecomment-1577626564

Both was made in the same machine.

Issue #94, opened 2020-10-06
Block based on ASN as well besides IP-Address/GEO-Location?


cbuijs:
Hi!

Seems that the GeoLite2 GEO-DB also provides the **ASN** (Autonomous System Number) for particular address-spaces. Would it be possible to block by ASN to addition of `cidr` and `location`?

It is actually very similar as `location` in usage.

For both `client-blocklist` and `response-blocklist-ip`.

For example, if I want to block anything of "Google", it would be looking like this:
```
[groups.response-blocklist-ip]
type = "response-blocklist-ip"
resolvers = ["internet-resolver"]
location-db = "/usr/share/GeoIP/GeoLite2-City.mmdb"
filter = true
blocklist-resolver = "nxdomain"
blocklist-format = "asn"
blocklist = [
    15169,
    16591,
    19527,
    36384,
    36385,
    36492,
    41264,
    43515,
    45566,
    139070,
    139190,
    394699,
    395973,
    396982,
]
```
Also ability to load lists via `source` etc.

ASN example Info: https://ipinfo.io/AS15169
IPASN Database: https://raw.githubusercontent.com/cbuijs/ipasn/master/ipasn-all.dat
ASN: https://en.wikipedia.org/wiki/Autonomous_system_(Internet)

-Chris.

folbricht:
This is doable. It seems to require a different database file, `GeoLite2-ASN` which isn't included by default.

cbuijs:
Ping!

When I check the DB directory, after a `geoipupdate`, it has these files:
```
ls -al /usr/share/GeoIP/
-rw-r--r--   1 root root  7824637 Aug  9 12:00 GeoLite2-ASN.mmdb
-rw-r--r--   1 root root 65858095 Aug  5 12:00 GeoLite2-City.mmdb
-rw-r--r--   1 root root  5590822 Aug  5 12:00 GeoLite2-Country.mmdb
```
Including the `GeoLite2-ASN.mmdb` one.

Tried to point `location-db` to it (no errors loading), but doesn't work as`response-blocklist-ip` only allows/accepts CIDR Prefixes in the `blocklist` and not ASN's.

folbricht:
Can you try the `issue-94` branch? Config example in https://github.com/folbricht/routedns/pull/330

cbuijs:
Tried it, but seems to do nothing... I suspect my local geolite is messed up. Need a bit more time, will test in the weekend.

folbricht:
[...]

PR #330, opened 2023-09-02
Implement GeoIP blocking based on ASN

folbricht:
Support for blocking of IPs from specific organizations (ASN). Requires `/usr/share/GeoIP/GeoLite2-ASN.mmdb` or similar to be present which may require a GeoLite2 account + running `geoipupdate` on Linux.

```toml
 [groups.cloudflare-blocklist]
type                = "response-blocklist-ip"
resolvers           = ["cloudflare-dot"]
blocklist-format    = "asn"
blocklist           = [
  "15169", # Google
]
```

Implements #94

### DIFF
--- a/asn-db.go
+++ b/asn-db.go
@@ -1,0 +1,96 @@
+package rdns
+
+import (
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+
+	"github.com/oschwald/maxminddb-golang"
+)
+
+// ASNDB holds blocklist rules based on ASN location. When an IP is queried,
+// its ASN is looked up in a database and the result is compared to the
+// blocklist rules.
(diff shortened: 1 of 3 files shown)


folbricht:
@cbuijs This one is a bit older. Do you think it's worth merging? I haven't done any testing with it for quite a while. For now I'm going to rebase it at least.

cbuijs:
> @cbuijs This one is a bit older. Do you think it's worth merging? I haven't done any testing with it for quite a while. For now I'm going to rebase it at least.

Yes, I use it :-). But I can do without as well. Not a biggie to drop it, or leave it unmerged.

folbricht:
If it works and is useful, may as well merge it

PR #333, opened 2023-09-03
Treat 'reset-after' time in random group as seconds

### LINKED ISSUE #332: 100% CPU usage when using fastest and random groups together, with resolvers and bootstrap address pointing to a unreachable network address.
This is my config that I use in my Linux PCs, it points to a DNSCrypt instance and many DoH HTTP/3 resolvers using fastest and random group schemes: https://paste.ee/p/foKdg

The address "10.8.56.2" is of a VPN that starts automatically after RouteDNS using systemd.

The problem is that RouteDNS uses 100% of CPU until the address "10.8.56.2" is reachable. Sometimes I use my PC with network turned off and this have been an issue.

I was able to reproduce this behavior, and I realized that if the bootstrap address has the option `local-address` and the network IP is not existent, an infinite loop of queries of bootstrap address is made: https://paste.ee/r/4g6me/0

This problem is happening with [master](url) branch.

@folbricht 

I modified my config to run as normal user, I think you'll be able to reproduce the problem too: https://paste.ee/r/U3y4A/0

folbricht:
The number in the config was treated as nanoseconds, not seconds as it should have been per documentation.

Fixes #332

### DIFF
--- a/cmd/routedns/main.go
+++ b/cmd/routedns/main.go
@@ -331,7 +331,7 @@ func instantiateGroup(id string, g group, resolvers map[string]rdns.Resolver) er
 		resolvers[id] = rdns.NewFailRotate(id, opt, gr...)
 	case "fail-back":
 		opt := rdns.FailBackOptions{
-			ResetAfter:    time.Duration(g.ResetAfter),
+			ResetAfter:    time.Duration(time.Duration(g.ResetAfter) * time.Second),
 			ServfailError: g.ServfailError,
 		}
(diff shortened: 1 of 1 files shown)


alpominth:
It was fixed: https://github.com/folbricht/routedns/issues/332#issuecomment-1705523726

It's ready to merge.

charlieporth1 (approved):

charlieporth1:
@folbricht I am merging this one for speedy reasons because it's rather crucial.

PR #338, opened 2023-09-11
added a null check for issue-337

charlieporth1:


folbricht:
This only addresses a symptom of the bug, not the cause. If you're able to reproduce it, can you find where it adds `nil` to the cache?

charlieporth1:
> This only addresses a symptom of the bug, not the cause. If you're able to reproduce it, can you find where it adds `nil` to the cache?
Would a null check [here](https://github.com/folbricht/routedns/blob/20545c9a53e3ba24542d4ba596e92a741636c482/lru-cache.go#L83C1-L83C23) be better? It is crashing on my server every 60 seconds which is inline with the gc



> This only addresses a symptom of the bug, not the cause. If you're able to reproduce it, can you find where it adds `nil` to the cache?

folbricht:
Let's find out where it comes from. Maybe add this

```go
func (b *memoryBackend) Store(query *dns.Msg, item *cacheAnswer) {
        if item == nil {
           panic("nil added to cache")
        }
	b.mu.Lock()
	b.lru.add(query, item)
	b.mu.Unlock()
}
```

That's going to give us a stacktrace of whatever is putting a nil answer into the cache.

PR #373, opened 2024-03-23
Add support for extended error codes in blocklist-v2

folbricht:
Supports extended errors for blocklists. The `text` field in the error message is a template that has access to the query and can use that to customize the message.

```toml
[groups.cloudflare-blocklist]
type              = "blocklist-v2"
resolvers         = ["cloudflare-dot"]
blocklist-format  = "domain"
edns0-ede = {code = 15, text = "Blocked {{ .Question }} with ID {{ .ID }} because reasons "} # Extended error code
blocklist         = [
  'evil.com',
]
```

`edns0-ede` will only be used if the blocklist actually blocks, not when it just spoofs response IPs.


Anuskuss:
Works but could this be extended to the other block types (i.e. `response-blocklist-ip`)?
Also that syntax looks clunky but if it's properly documented, I have no complains.

folbricht:
Just added the same to `response-blocklist-*` and `static-responder` (not tested). As for the syntax, would it be better to simplify to something like `{{ .Question }}` ? Since it's unlikely anything else is needed from the query.

Anuskuss:
> Just added the same to response-blocklist-*

Works now with `response-blocklist-ip` but I don't know the right keyword to get the IP address (if that's even possible).

> and static-responder (not tested).

Enabling `static-responder` (without anything) results in a segmentation fault.

> As for the syntax, would it be better to simplify

I'd suggest a bash-like syntax like
```
text = 'Blocked because $qname got caught by "${filter}" ($qtype)'
```
(`$filter` in my case would be the IP address)

Other (less useful) variables could be
- `$date` (current date in ISO 8601)
- `${env[]}` (array of enviroment variables)
- `$version` (vX.Y)
- `$mtime` (static = `mtime` of `config.toml`, dynamic = `Last-Modified` of document)
- `$id`

folbricht:
Fixed the nil pointer, missed that one earlier. As for the placeholders, I do want to keep using templating offered by the standard library, so can support syntax like `{{ .Question }}` and `{{ .ID }}`.

Anuskuss:
>  so can support syntax like `{{ .Question }}` and `{{ .ID }}`

This is fine by me. Like I said, I don't care how it looks as long as it's documented. As for the IP address, does that work yet? What's the keyword? Or would that not work with these "templates"?

folbricht:
Finally found a bit of time to work on this again. I simplified the way templates can access query values. Now it's possible to just use `"Blocked {{ .Question }} with ID {{ .ID }}"`. At this point, this only works for blocklists though, still working on how to get the static-responder to use it (in another PR). Wondering if it'd be better to make a new element like `template-responder` or `dynamic-responder`. I should be able to make it support the same template syntax as in the extended error strings.

PR #387, opened 2024-05-06
Enabling 0-RTT for QUIC/H3 clients

LeonardWalter:
Updated the DoQ and DoH QUIC client to enable 0-RTT based on the guide from: https://quic-go.net/docs/http3/client/#using-0-rtt

### DIFF
--- a/dohclient.go
+++ b/dohclient.go
@@ -181,7 +181,12 @@ func (d *DoHClient) ResolveGET(q *dns.Msg) (*dns.Msg, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), d.opt.QueryTimeout)
 	defer cancel()
 
-	req, err := http.NewRequestWithContext(ctx, "GET", u, nil)
+	method := http.MethodGet
+	if d.opt.Transport == "quic" {
+		method = http3.MethodGet0RTT
+	}
+
+	req, err := http.NewRequestWithContext(ctx, method, u, nil)
 	if err != nil {
 		d.metrics.err.Add("http", 1)
 		return nil, err
@@ -268,6 +273,9 @@ func dohQuicTransport(endpoint string, opt DoHClientOptions) (http.RoundTripper,
 	if err != nil {
 		return nil, err
 	}
+
+	// enable TLS session caching for session resumption and 0-RTT
+	tlsConfig.ClientSessionCache = tls.NewLRUClientSessionCache(100)
 	tlsConfig.ServerName = u.Hostname()
 	lAddr := net.IPv4zero
 	if opt.LocalAddr != nil {
--- a/dohclient.go
+++ b/dohclient.go
@@ -187,7 +187,7 @@ func (d *DoHClient) ResolveGET(q *dns.Msg) (*dns.Msg, error) {
 	defer cancel()
 
 	method := http.MethodGet
-	if d.opt.Use0RTT {
+	if d.opt.Use0RTT && d.opt.Transport == "quic" {
 		method = http3.MethodGet0RTT
 	}
 
--- a/dohclient.go
+++ b/dohclient.go
@@ -44,6 +44,8 @@ type DoHClientOptions struct {
 
 	// Optional dialer, e.g. proxy
 	Dialer Dialer
+
+	Use0RTT bool
 }
 
 // DoHClient is a DNS-over-HTTP resolver with support fot HTTP/2.
@@ -84,6 +86,9 @@ func NewDoHClient(id, endpoint string, opt DoHClientOptions) (*DoHClient, error)
 
 	if opt.Method == "" {
 		opt.Method = "POST"
+	}
+	if opt.Use0RTT && opt.Transport == "quic" {
+		opt.Method = "GET"
 	}
 	if opt.Method != "POST" && opt.Method != "GET" {
 		return nil, fmt.Errorf("unsupported method '%s'", opt.Method)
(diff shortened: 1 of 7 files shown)


dohclient.go
```diff
@@ -181,7 +181,12 @@ func (d *DoHClient) ResolveGET(q *dns.Msg) (*dns.Msg, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), d.opt.QueryTimeout)
 	defer cancel()
 
-	req, err := http.NewRequestWithContext(ctx, "GET", u, nil)
+	method := http.MethodGet
+	if d.opt.Transport == "quic" {
```
folbricht, line 185: Since 0-RTT affects the security-properties, what do you think about adding a flag to config like `enable-0rtt` or so to enable it. In the docs we could then link to https://datatracker.ietf.org/doc/html/rfc8446#section-8 so users know and can decide if they want it anyway.

folbricht (commented):
Much appreciated

LeonardWalter:
I updated the DoQ client to use DialEarly. The implementation looks a bit weird but others (SSH3 for example) are also using `net.ListenUDP(netString, nil)` for this purpose.
In my testbed with adguard's dnsproxy DoQ server, I can observe 0-RTT packets with DoQ.

doqclient.go
```diff
@@ -209,7 +216,21 @@ func (s *quicConnection) getStream(endpoint string, log *logrus.Entry) (quic.Str
...
+
+		// Resolve the UDP address for the endpoint
+		udpAddr, err := net.ResolveUDPAddr("udp", endpoint)
+		if err != nil {
+			log.WithError(err).Debug("couldn't resolve UDP address for endpoint [" + endpoint + "]")
+			return nil, err
+		}
+
+		// Use quic.DialEarly to attempt to use 0-RTT DNS queries for lower latency
+		s.EarlyConnection, err = quic.DialEarly(context.TODO(), s.udpConn, udpAddr, s.tlsConfig, s.config)
```
folbricht, line 233: I'm surprised that this change was even necessary. `quicDial` itself also calls `quic.DialEarly`. And it basically does the same thing.

Also there's a bit of a filehandle leak here when this fails. `s.udpConn` will not get closed. See the comment in https://github.com/folbricht/routedns/blob/master/dohclient.go#L427-L428

folbricht (commented):
Thanks a lot for this. As for docs, would you mind adding a note about the new flag to the documentation in https://github.com/folbricht/routedns/blob/master/doc/configuration.md#DNS-over-QUIC-Resolver ? If not I can do that later too though

doqclient.go
LeonardWalter: you are right, I tested it again and it works without the changes to use DialEarly. I will revert this

mattkeenan:
@LeonardWalter you mentioned in your commit message that there was a bug in the previous doqclient code; what was the bug?

LeonardWalter:
The problem was what Frank mentioned with the updConn not being closed.
If the query failed due to timeout or if the server rejects the query, the udpConn would remain open for some time causing new queries to fail.

mattkeenan:
@LeonardWalter to be clear it wasn't a criticism of the commit, i wasn't sure i understood it correctly; apologies. also, so i understand correctly, dohclient.go will do the correct thing for http2 / TCP?

dohclient.go
```diff
@@ -181,7 +186,12 @@ func (d *DoHClient) ResolveGET(q *dns.Msg) (*dns.Msg, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), d.opt.QueryTimeout)
 	defer cancel()
 
-	req, err := http.NewRequestWithContext(ctx, "GET", u, nil)
+	method := http.MethodGet
+	if d.opt.Use0RTT {
```
folbricht, line 190:
```suggestion
	if d.opt.Use0RTT && opt.Transport == "quic" {
```
Should we add this? To handle a mis-configuration where Use0RTT is used but it's not actually quic, but http2

LeonardWalter:
@mattkeenan 
It is good that you mention this. The changes don't affect DoH much, but introducing the session cache (`tlsConfig.ClientSessionCache = tls.NewLRUClientSessionCache(100)`) does change the behavior.
Session resumption should also be used for HTTP/2, but I have not tested it.

I looked at the DoH RFC
>    TLS-based implementations often achieve better handshake performance
>    through the use of some form of session resumption mechanism, such as
>    [Section 2.2 of [RFC8446]](https://datatracker.ietf.org/doc/html/rfc8446#section-2.2).  Session resumption creates trivial
>    mechanisms for a server to correlate TLS connections together.

AFAIK there are countermeasures to this in TLS1.3 but don't quote me on this I am no expert. Other DoH implementations also it.
From the DNS over QUIC RFC but this also counts for DoH when using session resumption/pre shared keys
>    Session resumption and 0-RTT data transmission create privacy risks
>    detailed in Sections 7.1 and 7.2.  The following recommendations are
>    meant to reduce the privacy risks while enjoying the performance
>    benefits of 0-RTT data, subject to the restrictions specified in
>    Section 4.5.
> 
>    Clients SHOULD use resumption tickets only once, as specified in
>    Appendix C.4 of [RFC8446].  By default, clients SHOULD NOT use
>    session resumption if the client's connectivity has changed.

folbricht:
Server side? Is it needed? Seems fine to support it in all cases

folbricht (approved):

Issue #398, opened 2024-06-26
Ellaborate "Client Blocklist" to use the EDNS Mac-address option (65001)


cbuijs:
For `client-blocklist` it would be nice to also be able to filter baed on on mac-address.

Either straight from the query (hard to do and not uniform on different platforms), or just from EDNS Option 65001.

We discussed this earlier in #105 already, but I think supporting macs in a list utilising EDNS Option 65001 is quite accepted standard nowadays (see PowerDNS, DNSDist and DNSMasq for example).

folbricht:
This was more complex that I thought but there's an implementation on `issue-398` branch. Can you give it a try? I added a unit test but haven't been able to test fully end-to-end.

PR #401, opened 2024-07-07
Support blocking based on MAC-address

folbricht:
Adds a new blocklist format "mac" which can be used to filter based on MAC addresses in EDNS0 option 65001

```toml
[groups.cloudflare-blocklist]
type             = "blocklist-v2"
resolvers        = ["cloudflare-dot"]
blocklist-format = "mac"
blocklist        = [ # Define the MAC addresses to be blocked, these are expected to be supplied by the client in EDNS0 option 65001
  '01:23:45:ab:cd:ef',
  '01:01:01:ff:ff:ff',
]
```
Implements #398


cbuijs:
I think this is fine to merge. No issues since I use it.

Issue #409, opened 2024-08-31
Query logs


marcelloinfoweb:
How is it possible to show a log of requests in the terminal in real time?

folbricht:
I wanted to add this at some point anyway. Would you be able to try out the `issue-409` branch? See https://github.com/folbricht/routedns/pull/410 for an example.

cbuijs:
You can already do so with the syslog stuff?

folbricht:
True, it's been a while and I completely forgot about that, doh. The new one is a bit easier to use but this does look unnecessary. Should probably keep just syslog for logging and not merge #410

cbuijs:
No worries. There is a case for both anyway ;-)

marcelloinfoweb:
> I wanted to add this at some point anyway. Would you be able to try out the `issue-409` branch? See #410 for an example.

I depend on a binary, I can't compile it.

marcelloinfoweb:
Error routedns[7180]: Error: unsupported group type 'query-log' for group 'query-log'

> I wanted to add this at some point anyway. Would you be able to try out the `issue-409` branch? See #410 for an example.

Error: routedns[7180]: Error: unsupported group type 'query-log' for group 'query-log'

folbricht:
That won't work for you unless you can compile it from the branch. If you can't compile it, you could try out `syslog` instead

marcelloinfoweb:
> That won't work for you unless you can compile it from the branch. If you can't compile it, you could try out `syslog` instead. There's an example in https://github.com/folbricht/routedns/blob/master/cmd/routedns/example-config/syslog.toml

I did it, I compiled the branch and now I have a log, but it doesn't show the blocked and/or released lists.

folbricht:
`syslog` does log responses as well if that's what you're looking for.

cbuijs:
If you want to log the blocks/allows. You need to start with `-l5` or `-l6` as parameter.

I start RouteDNS as follows and have all the logging in a file (`routedns.log`):
`./routedns -l6 routedns.toml >> routedns.log 2>&1 & disown`

Or get everything into syslog using logger:
`./routedns -l6 routedns.toml 2>&1 | /usr/bin/gawk '{print substr($0, index($0, $3))}' | /usr/bin/logger -t "ROUTEDNS" & disown`

You can play with the log levels to get more/less logging. I think from `-l5` on, blocking is logged.

Above examples make RouteDNS run in the background and it stays there running when you log off.

[...]

PR #410, opened 2024-09-01
Implement 'query-log'

folbricht:
Implements #409 

Adds a new `query-log` element that can log to file or STDOUT.

```toml
[groups.query-log]
type   = "query-log"
output-file = "/tmp/query.log" # Comment out to log to STDOUT
```


cbuijs:
Could we add something so that the EDNS-ECS address is display in the log as well, or instead of the client-ip?

folbricht:
Can certainly be done. To facilitate more values getting added, it's probably best to also switch to structured logging rather than using more columns.

folbricht:
Added that, the output looks like this now:
```
time=2024-09-29T13:17:12.580+02:00 source-ip=127.0.0.1 question-name=google.com. question-class=IN question-type=A
```

Wondering if it'd be better to output these as JSON, or at least offer that as an option

cbuijs:
Having plain-text and json as (both) options would be great. Tailing a json file is a bit tricky if you want to monitor (where I use it for).

folbricht:
Added support for JSON output (optional). This produces JSONL which can be piped into `jq .`

cbuijs:
I think this is fine to merge. No issues at all last months when using it all the time.

folbricht:
Thanks for testing

PR #414, opened 2024-11-19
Add support for 0rtt connection resumption to DoH and DoQ listeners. 

LeonardWalter:
Add support for 0rtt connection resumption to DoH and DoQ listeners.

### DIFF
--- a/doqlistener.go
+++ b/doqlistener.go
@@ -128,7 +128,6 @@ func (s DoQListener) handleConnection(connection quic.Connection) {
 	for {
 		stream, err := connection.AcceptStream(context.Background())
 		if err != nil {
-			log.Trace("Failed to create quic stream, closing connection")
 			break
 		}
 		log.WithField("stream", stream.StreamID()).Trace("opening stream")
--- a/doqlistener.go
+++ b/doqlistener.go
@@ -20,7 +20,7 @@ type DoQListener struct {
 	addr    string
 	r       Resolver
 	opt     DoQListenerOptions
-	ln      *quic.Listener
+	ln      *quic.EarlyListener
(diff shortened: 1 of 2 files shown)


doqlistener.go
```diff
@@ -128,31 +126,31 @@ func (s DoQListener) handleConnection(connection quic.Connection) {
...
 	// DNS over QUIC uses one stream per query/response.
-	defer stream.Close()
 	s.metrics.stream.Add(1)
 
 	// DoQ requires a length prefix, like TCP
 	var length uint16
 	if err := binary.Read(stream, binary.BigEndian, &length); err != nil {
 		s.metrics.err.Add("read", 1)
 		log.WithError(err).Error("failed to read query")
+		CloseConn(connection, log)
```
folbricht, line 153: Not a fan of closing the connection in multiple places, including from within the stream. One effect of doing it here will be that `connection.AcceptStream(ctx)` above will return an error, which will then call `CloseConn` a second time. It should write `"closing connection"` twice in the logs too.

The prior implementation only closed it in one place. Is that not working? If a stream fails, I'd expect `AcceptStream()` to also fail, which then closes the connection.

folbricht (commented):
Thank you. Let's make sure that closing the connection in multiple places is actually necessary.

doqlistener.go
LeonardWalter: I agree. I did remove the changes and updated the PR

LeonardWalter:
There is currently still a problem when the server restarts and the client still holds an old session ticket. I think the client should automatically retry without the session ticket after it failed to (re)open the connection. Will work on this now

PR #419, opened 2024-12-21
Bunch of DoH fixes bundled together

LeonardWalter:
I hope this is okay to do a PR like this. I noticed a couple issues with DoH and I just bundled them together here.
ioutil and hostname are replaced/removed because of feedback from the compiler.
Then I noticed that 0RTT is used even if the flag is not set. Furthermore quic-go appears to even use 0RTT with HTTP POST requests when a custom h3 transport is used. I am not sure if this is a known bug but to me it seems super  strange that this is not handled.
To restrict this aggressive use of 0-RTT I just disabled the session token cache when the flag is not enabled.

The other issue I found with the DoH listener, is that incoming requests are handled regardless of the specified path. This means that the current DoH listener would also respond to DoH messages received on any path like `example.com/whatever` .
In this fix I added a statement to drop requests not on `/dns-query`. Alternatively the handler `http.HandlerFunc(l.dohHandler)` could be replaced with something like `http.HandleFunc("/dns-query", l.dohHandler)`

Additionally, I added a check for `content-type "application/dns-message"`

### DIFF
--- a/dohclient.go
+++ b/dohclient.go
@@ -305,7 +305,7 @@ func dohQuicTransport(endpoint string, opt DoHClientOptions) (http.RoundTripper,
 	}
 
 	dialer := func(ctx context.Context, addr string, tlsConfig *tls.Config, config *quic.Config) (quic.EarlyConnection, error) {
-		return newQuicConnection(u.Hostname(), addr, lAddr, tlsConfig, config)
+		return newQuicConnection(u.Hostname(), addr, lAddr, tlsConfig, config, opt.Use0RTT)
 	}
 	if opt.BootstrapAddr != "" {
 		dialer = func(ctx context.Context, addr string, tlsConfig *tls.Config, config *quic.Config) (quic.EarlyConnection, error) {
@@ -314,7 +314,7 @@ func dohQuicTransport(endpoint string, opt DoHClientOptions) (http.RoundTripper,
 				return nil, err
 			}
 			addr = net.JoinHostPort(opt.BootstrapAddr, port)
-			return newQuicConnection(u.Hostname(), addr, lAddr, tlsConfig, config)
+			return newQuicConnection(u.Hostname(), addr, lAddr, tlsConfig, config, opt.Use0RTT)
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


dohclient.go
```diff
@@ -287,6 +287,11 @@ func dohQuicTransport(endpoint string, opt DoHClientOptions) (http.RoundTripper,
 		lAddr = opt.LocalAddr
 	}
 
+	qConf := &quic.Config{}
+	if opt.Use0RTT && opt.Method == http.MethodGet {
```
folbricht, line 291: Are you sure about this change? Doesn't it make sense for there to always be a token store, regardless of 0RTT?

dohlistener.go
```diff
@@ -155,6 +155,11 @@ func (s *DoHListener) String() string {
 }
 
 func (s *DoHListener) dohHandler(w http.ResponseWriter, r *http.Request) {
+	if r.URL.Path != "/dns-query" {
```
folbricht, line 158: Would it be worth mentioning the path as well as the content-type in the [docs](https://github.com/folbricht/routedns/blob/master/doc/configuration.md#dns-over-https)?

dohclient.go
LeonardWalter: I am not 100% sure about this. Alternatively `SessionTicketsDisabled = true` could be used but that does basically the same.
I have not found an alternative way. From my testing experience, the use of 0RTT with a custom http.transport seemed very aggressive. 
Ideally I would expect a regular 1RTT opening when using anything but http.MethodGet0RTT but this seems to not be the case.

dohlistener.go
LeonardWalter: I am reconsidering if this should be included. I looked at the RFC but it seems like the path `/dns-query` is not a MUST, rather a norm. 

I can see that this could maybe cause issues, since with routedns users can define different paths than `/dns-query` on the DoH client.

folbricht: yeah, it may be best to avoid breaking existing setups that use a non-standard path. One thing we could do is make the path configurable, and if it's not set, then all paths are allowed. This would not impact existing configs while still allowing someone to enforce a specific path. And it's fits nicely with routedns' overall configurability

LeonardWalter:
I removed the commit with the DoH listener changes from the PR

dohclient.go
LeonardWalter: The alternative would be to switch between using DialEarly and Dial in the quicDial function if 0RTT is enabled or not. But this would require bigger changes since dialEarly returns a quic.EarlyConnection and then the other functions are also required to support both quic.EarlyConnection and quic.Connection

LeonardWalter: Actually it was not that complicated. Here is what I came up with 56224a1142f62be487994f77f52545ff6fc46d4c

folbricht: That's looking pretty good, perhaps it's possible to remove the wrapper as well. Let's integrate that into this PR if possible

PR #422, opened 2024-12-31
move from logrus to slog

mosajjal:
Hi, 

not sure if this is something you're keen for. The reason behind this for my own use-cases was the fact that logrus doesn't compile for wasm/wasi. also reduces dependencies in general.

this refactor is a bit messy and inconsistent but it works.


folbricht:
Only had a quick look but this is promising. Thank you. Will check it some more and merge when I'm back after the holidays.

pipeline.go
```diff
@@ -153,18 +154,18 @@ func (c *Pipeline) start() {
 						// got a bad packet (like a truncated one for example).
 						if a == nil {
 							c.metrics.err.Add("read", 1)
-							log.WithError(err).Error("read failed")
+							Log.Error("read failed", "error", err)
```
folbricht, line 157:
```suggestion
							log.Error("read failed", "error", err)
```

resolver_test.go
```diff
@@ -8,7 +8,8 @@ import (
 
 func init() {
 	// Silence the logger while running tests
-	Log.SetLevel(0)
+	// TODO:fix
```
folbricht, line 11: Implementing this should be quite straightforward. Either replace the logger with one writing to io.Discard or set the level so high that it's above "error" level

response-blocklist-name.go
```diff
@@ -104,15 +104,15 @@ func (r *ResponseBlocklistName) blockIfMatch(query, answer *dns.Msg, ci ClientIn
...
 				if r.BlocklistResolver != nil {
-					log.WithField("resolver", r.BlocklistResolver).Debug("blocklist match, forwarding to blocklist-resolver")
+					log.With("resolver", r.BlocklistResolver).Debug("blocklist match, forwarding to blocklist-resolver")
 					return r.BlocklistResolver.Resolve(query, ci)
 				}
 				log.Debug("blocking response")
 				answer = nxdomain(query)
 				if err := r.EDNS0EDETemplate.Apply(answer, EDNS0EDEInput{query, rule}); err != nil {
-					log.WithError(err).Error("failed to apply edns0ede template")
+					Log.Error("failed to apply edns0ede template", "error", err)
```
folbricht, line 115:
```suggestion
					log.Error("failed to apply edns0ede template", "error", err)
```

pipeline.go
```diff
@@ -83,14 +83,14 @@ func (c *Pipeline) start() {
...
+	log := Log.With("addr", c.addr)
 	for req := range c.requests { // Lazy connection. Only open a real connection if there's a request
 		done := make(chan struct{})
-		log.Trace("opening connection")
+		log.Debug("opening connection")
 		conn, err := c.client.Dial(c.addr)
 		if err != nil {
 			c.metrics.err.Add("open", 1)
-			log.WithError(err).Error("failed to open connection")
+			Log.Error("failed to open connection", "error", err)
```
folbricht, line 93:
```suggestion
			Log.Error("failed to open connection", "error", err)
```

folbricht (approved):

folbricht:
Thank you

PR #426, opened 2025-01-11
Re-add support for Oblivious DNS

folbricht:
Adds ODNS support again after it had to be removed due to a broken dependency in https://github.com/folbricht/routedns/pull/425

@LeonardWalter Any idea how to deal with the indirect dep on `git.schwanenlied.me/yawning/x448.git`? Doesn't look like that's ever coming back online.


LeonardWalter:
Yes, I did check out the two repos cloudflare/odoh-go the forked sthorne/odoh-go. sthorne/odoh-go is actually a couple commits behind Cloudflare's version. It still uses the git.schwanenlied.me/yawning/x448.git implementation of x448. The official odoh-go repo already replaced this with their own x448 implementation that is part of cloudflare/circl. The cloudflare/odoh-go repo seems to have all the issues fixed, the problem is that for some reason, go is by default not using the latest version of the code from github.
Also there does not seem to be a more recent version available:
 ```
$ go list -m -versions github.com/cloudflare/odoh-go
>   github.com/cloudflare/odoh-go v0.1.2 v0.1.3 v0.1.4 v0.1.5 v0.1.6 v1.0.0
```

So I had to add it manually by using the commit hash from the latest commit on  github.com/cloudflare/odoh-go

```
$ go get github.com/cloudflare/odoh-go@f39fa019b017510690599e895c20cd02ae138742
>   go: downloading github.com/cloudflare/odoh-go v1.0.1-0.20230926114050-f39fa019b017
>   go: added github.com/cloudflare/odoh-go v1.0.1-0.20230926114050-f39fa019b017
```

No clue why this is needed but I am pretty sure that this does now fix all the dependency issues related to ODoH.

I will create a PR against the re-add-odoh branch

PR #429, opened 2025-01-13
Dnssec

LeonardWalter:
This is to start the process of integrating a DNSSEC validator to routedns. 
This implementation is inspired by [uw-ictd/DNSSEC-Validator](https://github.com/uw-ictd/DNSSEC-Validator) and [peterzen/goresolver](https://github.com/peterzen/goresolver), but adds concurrency and is adapted to work with the other routedns modules.

Currently, only validated DNS responses are forwarded. DNS responses that are not singed or have an invalid signature are dropped.

Example config file:
```
[resolvers.my-doh]
protocol = "doh"
address = "https://d.adguard-dns.com/dns-query{?dns}"

[groups.dnssec]
type = "dnssec"
resolvers = ["my-doh"]

[listeners.local-udp]
address = "127.0.0.1:9009"
protocol = "udp"
resolver = "dnssec"
```


dnssec-backend.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	q.IsEdns0().SetDo()
+	return q
+}
+
+func newQuery(qname string, qtype uint16) *dns.Msg {
+	dnsMessage := &dns.Msg{
+		MsgHdr: dns.MsgHdr{
+			RecursionDesired: true,
+		},
```
folbricht, line 94: Technically you don't need this as `SetQuestion()` below does it already (see https://github.com/miekg/dns/blob/v1.1.63/defaults.go#L35).

dnssec-backend.go
```diff
@@ -0,0 +1,313 @@
...
+
+	var g errgroup.Group
+	g.Go(func() error {
+		var err error
+		signedZone.Dnskey, err = getRRset(domainName, dns.TypeDNSKEY, resolver, ci)
+		if err != nil {
+			return err
+		}
+
+		for _, rr := range signedZone.Dnskey.RrSet {
```
folbricht, line 159: `getRRSet()` can return `nil, nil` which would trigger a panic here

dnssec-backend.go
```diff
@@ -0,0 +1,313 @@
...
+	q := newQuery(qname, qtype)
+	r, err := doQuery(q, resolver, ci)
+	if err != nil || r == nil {
+		return nil, err
+	}
+	return extractRRset(r)
+}
+
+func extractRRset(r *dns.Msg) (*RRSet, error) {
+	result := &RRSet{RrSet: make([]dns.RR, 0)}
```
folbricht, line 111:
```suggestion
	result := &RRSet{}
```
No need to initialize an empty list, you can read from and append to a `nil` list.

dnssec-backend.go
```diff
@@ -0,0 +1,313 @@
...
+	r, err := resolver.Resolve(q, ci)
+	if err != nil {
+		return nil, err
+	}
+	if r.Rcode == dns.RcodeNameError {
+		Log.Info("no such domain", "info", qName(r))
+		return nil, ErrNoResult
+	}
+	if r == nil || r.Rcode == dns.RcodeSuccess {
+		return r, err
```
folbricht, line 138:
```suggestion
		return r, nil
```
`err` has already been checked and must be `nil` here.

dnssec-backend.go
```diff
@@ -0,0 +1,313 @@
...
+
+		for _, rr := range signedZone.Dnskey.RrSet {
+			defer func() {
+				if r := recover(); r != nil {
+					Log.Warn("panic occurred",
+						slog.String("domainName", domainName),
+						slog.Any("panic", r),
+					)
+				}
+			}()
```
folbricht, line 167: Do we really need this? Seems more like we don't trust our own code to handle all nil pointers correctly.

dnssec-backend.go
```diff
@@ -0,0 +1,313 @@
...
+	g.Go(func() error {
+		var err error
+		signedZone.Ds, err = getRRset(domainName, dns.TypeDS, resolver, ci)
+		if err != nil {
+			Log.Error("DS query failed",
+				slog.String("domainName", domainName),
+				"error", err,
+			)
+
+			return err
```
folbricht, line 182: Rather than logging from within, might be better to extend the error and pass it back to the caller to handle. Something like

```go
return fmt.Errorf("DS query failed for %q: %w", domainName, err)
```

dnssec-backend.go
```diff
@@ -0,0 +1,313 @@
...
+				authChain.DelegationChain[index-1].ParentZone = delegation
+			}
+			return nil
+		})
+	}
+
+	if err := g.Wait(); err != nil {
+		return err
+	}
+	return nil
```
folbricht, line 219:
```suggestion
	return g.Wait()
```

dnssec-backend.go
```diff
@@ -0,0 +1,313 @@
...
+	if err != nil {
+		return ErrInvalidRRsig
+	}
+
+	for _, signedZone := range authChain.DelegationChain {
+		defer func() {
+			if err := recover(); err != nil {
+				Log.Error("AuthChain panic occurred", "error", err)
+			}
+		}()
```
folbricht, line 243: Again, there should be a better way to deal with panics, i.e. avoid them at all.

folbricht (commented):
What happens when a domain isn't signed? Will this fail? I'm wondering if there should be a flag or so to switch into a more permissive mode where DNSSEC responses are validated, but responses for unsigned domains are passed through.

folbricht:
Is there anything else you're planning to add to this PR?

Overall my main concern is around calling it DNSSEC when it really is just a partial implementation. There are several more safeguards in the spec that aren't really handled by this implementation. It gets complex fast, which is also why I haven't tried it myself yet. Including, but probably not limited to:
- Root key management (didn't notice anything around that)
- Verifying NXDOMAIN responses, there is a dnssec-way of proving the domain doesn't exist vs a fake response
- Pretty sure there is also a way to prove that a domain isn't signed by checking the parent, rather than assuming it's not when the keys are missing in the response.

Perhaps we should call out that this is just a partial/experimental implementation? There are quite a few weaknesses in it and users shouldn't fully rely on it.

LeonardWalter:
I am currently on vacation, so I won't be able to work on this until the end of the month. I should have some time then to dive deeper into DNSSEC and extend this PR.

You can decide if you want to merge it already, but I agree with your point that some additional information should be included.

LeonardWalter:
I added support for loading in trust-anchor root-keys from a XML file. (Currently it only supports loading in the root zone "." )

And in the latest commit I added support for NSEC and NSEC3 Proof of Non-Existence. I am not confident about my code here, but it seemed to work for all domains I tested so far.

I couldn't find anything on how to prove unsigned domains are to be unsigned.

LeonardWalter:
I haven't really worked on this since last year and i can see that you merged you branch already so im closing this :)

PR #430, opened 2025-01-13
ODoH listener

LeonardWalter:
This PR adds an Oblivious DNS over HTTPS listener with proxy and target support.

### DIFF
--- a/odohlistener.go
+++ b/odohlistener.go
@@ -6,6 +6,7 @@ import (
 	"crypto/tls"
 	"encoding/hex"
 	"errors"
+	"fmt"
 	"io"
 	"log/slog"
 	"net/http"
@@ -30,6 +31,7 @@ type ODoHListener struct {
 	addr        string
 	proxyClient *http.Client // Client for proxy forwarding if acting as ODoH proxy
 	doh         *DoHListener
+	config      []byte
 
 	r           Resolver // Forwarding DNS queries if acting as ODoH Target
 	opt         ODoHListenerOptions
@@ -51,20 +53,10 @@ var _ Listener = &ODoHListener{}
 func NewODoHListener(id, addr string, opt ODoHListenerOptions, resolver Resolver) (*ODoHListener, error) {
 	keyPair, err := getKeyPair(opt)
 	if err != nil {
-		Log.Error("Failed to generate HPKE key pair", "error", err)
-		return nil, err
+		return nil, fmt.Errorf("failed to generate HPKE key pair: %w", err)
 	}
 
-	dohOpt := DoHListenerOptions{
-		TLSConfig: opt.TLSConfig,
-		isChild:   true,
-	}
-	dohListen, err := NewDoHListener(id, addr, dohOpt, resolver)
-	if err != nil {
-		Log.Error("Failed to spawn DoH listener", "error", err)
-		return nil, err
-	}
-
+	configSet := odoh.CreateObliviousDoHConfigs([]odoh.ObliviousDoHConfig{keyPair.Config})
 	l := &ODoHListener{
 		id:          id,
 		addr:        addr,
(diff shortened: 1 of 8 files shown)


odohlistener.go
```diff
@@ -0,0 +1,250 @@
...
+
+func (s *ODoHListener) Stop() error {
+	return s.doh.Stop()
+}
+
+func (s *ODoHListener) ODoHproxyHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "POST" {
+		http.Error(w, "only POST allowed", http.StatusMethodNotAllowed)
+		return
+	}
```
folbricht, line 104: A further advantage of using a `*http.ServeMux` is that you can specify the HTTP method it should handle in the pattern, making this check unnecessary. See https://pkg.go.dev/net/http#ServeMux

odohlistener.go
```diff
@@ -0,0 +1,250 @@
...
+		http.Error(w, "no targethost specified", http.StatusBadRequest)
+		return
+	}
+	path := r.URL.Query().Get("targetpath")
+	if path == "" {
+		http.Error(w, "no targetpath specified", http.StatusBadRequest)
+		return
+	}
+
+	defer r.Body.Close()
```
folbricht, line 117:
```suggestion
```
No need to close the request body, that's only needed for a response body.

odohlistener.go
```diff
@@ -0,0 +1,250 @@
...
+	b, err := io.ReadAll(r.Body)
```
folbricht, line 118: This would allow an attacker to cause an OOM of the service quite easily. We should define a limit (4k?) and use a io.LimitReader or such to protect the process. Then just close the connection if that limit is reached. We could go even further and stream the request through the proxy to the target without reading it all into memory first, but a small limit should do for now.

odohlistener.go
```diff
@@ -0,0 +1,250 @@
...
+	if err != nil {
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		return
+	}
+	if response.StatusCode != 200 {
+		http.Error(w, http.StatusText(response.StatusCode), response.StatusCode)
+		return
+	}
+
+	defer response.Body.Close()
```
folbricht, line 138: This needs to be moved up, above the status-code check to avoid a leak.

odohlistener.go
```diff
@@ -0,0 +1,250 @@
...
+	rb, err := io.ReadAll(response.Body)
```
folbricht, line 139: Same issue as above with this getting read fully into memory. The fix trivial in this case though. Just stream the response from the target through to the client like so:
```go
w.Header().Set("Content-Type", contentType) // Probably want to use the Content-Type received from the target

io.Copy(w, response.Body)
```

odohlistener.go
```diff
@@ -0,0 +1,250 @@
...
+	if err != nil {
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		return
+	}
+	w.Header().Set("Content-Type", contentType)
+	w.Write(rb)
+}
+
+func forwardProxyRequest(client *http.Client, targethost string, targetPath string, body []byte, contentType string) (*http.Response, error) {
+	targetURL := "https://" + targethost + targetPath
```
folbricht, line 149: This does look risky, and if got flagged by automatic code-scanning below. I don't remember how abuse is prevented in the RFC (will have to read it again). Maybe there's supposed to be an allowlist of target hosts? At least we should do more that trust all values and just concatenating into a URL. Perhaps:
- Validate the host is a valid hostname
- hostname, path length checks (limits)
- check against an allowlist?
- check the path to see if it's valid
- Build the URL with an appropriate library (https://pkg.go.dev/net/url) rather than string
operations
- ...

odohlistener.go
```diff
@@ -0,0 +1,250 @@
...
+			http.Error(w, "only contentType oblivious-dns-message allowed", http.StatusMethodNotAllowed)
+			return
+		}
+	}
+
+	if qHeader != ODOH_CONTENT_TYPE {
+		http.Error(w, "only contentType oblivious-dns-message allowed", http.StatusMethodNotAllowed)
+		return
+	}
+	defer r.Body.Close()
```
folbricht, line 177:
```suggestion
```

odohlistener.go
```diff
@@ -0,0 +1,250 @@
...
+	b, err := io.ReadAll(r.Body)
```
folbricht, line 178: io.LimitReader

odohlistener.go
```diff
@@ -0,0 +1,250 @@
...
+	p, err := a.Pack()
+	if err != nil {
+		Log.Error("failed to encode response", "error", err)
+		return
+	}
+
+	response := odoh.CreateObliviousDNSResponse(p, 0)
+	obliviousResponse, err := responseContext.EncryptResponse(response)
+	if err != nil {
+		http.Error(w, "failed to create oblivious response", http.StatusBadRequest)
```
folbricht, line 217: If we fail to encrypt the response, is that a BadRequest or an InternalServerError? Maybe the client provided a bad key?

odohlistener.go
```diff
@@ -0,0 +1,250 @@
...
+	w.Write(configs.Marshal())
+}
+
+func getKeyPair(opt ODoHListenerOptions) (odoh.ObliviousDoHKeyPair, error) {
+	var seed []byte
+	var err error
+	if opt.KeySeed != "" {
+		seed, err = hex.DecodeString(opt.KeySeed)
+		if err != nil {
+			Log.Error("Failed to read key seed", "error", err)
```
folbricht, line 241:
```suggestion
			return nil, fmt.Errorf("failed to read key seed: %w", err)
```

odohlistener.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+	} else {
+		seed = make([]byte, defaultSeedLength)
+		if _, err := rand.Read(seed); err != nil {
+			Log.Error("Failed to generate random seed", "error", err)
```
folbricht, line 246:
```suggestion
			return nil, fmt.Errorf("Failed to generate random seed: %w", err)
```

LeonardWalter:
Implemented all of the feedback except the security related comments with the proxy handler. Want to add this in an extra commit soon.

LeonardWalter:
Sorry for the messy forced update, that was a mistake.

Here is a list of the changes:
- added "dual" mode as option and made "target" the default mode
- switched to using http.ServeMux for ODoH listener and also the DoH listener - should work the same now after I switched back to using universal "/" handler on DoH listener (939e33d93d4963fd04df1159d26008e4d99bc52f).
- changed error handling 
- changed to io.LimitReader and io.Copy 
- used more standard HTTP Constants

For the proxy security warnings, I think the bot only had problems with the user specified "content-type" field. 
Here I made the switch to enforce ODOH_CONTENT_TYPE as this seems reasonable enough. 
Also added a RFC 1123 hostname regex and some length checks.

folbricht:
It's looking pretty good. Wrt the client connection pool, I don't think you even need to do anything there. There's already a shared http.Client in use for target connections, that has it's own pool and re-uses connections (assuming you read the whole response and close the response.Body). You may not need https://github.com/folbricht/routedns/commit/5edce156bf63366fa4b8b4f8828a339d50bec448 at all

folbricht:
As per https://pkg.go.dev/net/http#Client
> The [Client.Transport] typically has internal state (cached TCP connections), so Clients should be reused instead of created as needed. Clients are safe for concurrent use by multiple goroutines.

LeonardWalter:
I did need it for HTTP/3 though, that is why I implemented it.

Issue #433, opened 2025-02-03
`--log-level` doesn't work anymore


Anuskuss:
I run RouteDNS as a systemd service with `--log-level 1` because I really don't care about lines like these
```
time=2025-02-03T23:36:16.857+01:00 level=ERROR msg="failed to resolve" id=udp client=127.0.0.1 qname=###. protocol=udp addr=:5301 error="query for '###.' timed out"
```
flooding my journal. But now not even `--log-level 0` hides those.

folbricht:
Yes, I see the issue. This broke during the move to `slog` from `logrus`. Previously we used these levels https://pkg.go.dev/github.com/sirupsen/logrus#Level with 0 == Panic only. But since logrus doesn't have the same, we're mapping it like this https://github.com/folbricht/routedns/blob/master/cmd/routedns/main.go#L86-L98

I won't be able to fix it this week, but feel free to open a PR to get it close to what it was before.

Anuskuss:
So since `slog` seems to only have 4 levels (technically [they say](https://go.dev/blog/slog) that the level is an integer and that you can have more but 4 is enough IMO) and there's nothing less verbose than `ERROR` I'd say that line mentioned above shoud be increased to `WARN`:

```diff
--- a/cmd/routedns/main.go
+++ b/cmd/routedns/main.go
@@ -53,7 +53,7 @@ arguments.
                SilenceUsage: true,
        }
 
-       cmd.Flags().Uint32VarP(&opt.logLevel, "log-level", "l", 4, "log level; 0=None .. 6=Trace")
+       cmd.Flags().Uint32VarP(&opt.logLevel, "log-level", "l", 4, "log level; 0=None 1=Error 2=Warn 3=Info 4=Debug")
        cmd.Flags().BoolVarP(&opt.version, "version", "v", false, "Prints code version string")
 
        if err := cmd.Execute(); err != nil {
@@ -86,12 +86,14 @@ func start(opt options, args []string) error {
        var level slog.Level
        switch opt.logLevel {
        case 0:
+               // TODO: None
+       case 1:
                level = slog.LevelError
-       case 1, 2:
+       case 2:
                level = slog.LevelWarn
-       case 3, 4:
+       case 3:
                level = slog.LevelInfo
-       case 5, 6:
+       case 4:
                level = slog.LevelDebug
        default:
                level = slog.LevelInfo
--- a/dnslistener.go
+++ b/dnslistener.go
[...]

PR #435, opened 2025-02-09
Re-align log-levels with logrus

folbricht:
https://github.com/folbricht/routedns/pull/422 switched from logrus to slog. It slightly changed the meaning of log-levels which caused https://github.com/folbricht/routedns/issues/433

### DIFF
--- a/cmd/routedns/main.go
+++ b/cmd/routedns/main.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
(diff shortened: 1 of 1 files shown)


Anuskuss:
IMO `0` should be disabled and `1` should be `Error`. Also don't forget to update
https://github.com/folbricht/routedns/blob/383193dd987c696ec6c10d05c2ba6ae68f75153e/cmd/routedns/main.go#L57

Issue #443, opened 2025-03-20
Read/write raw responses


Anuskuss:
I think It would be pretty useful if you could read/write "raw" data from/into responses.

1. Read data
```toml
[groups.nocache]
type      = "ttl-modifier"
ttl-max   = 1 # 0 doesnt seem to work yet
resolvers = [ "cloudflare" ] # this doesnt run if called by "dnsflag"

[group.dnsflag]
type      = "raw-read"
resolvers = [ "cloudflare" ]                           # get the response
match     = { dns-flags = [ "QR", "RD", "RA", "AD" ] } # if: response contains AD
action    = [ "nocache" ]                              # then: do this
```

2. Write data
```toml
[groups.ednsflag]
type      = "raw-write"
resolvers = [ "cloudflare" ]          # get the response
add       = { edns-flags = [ "DO" ] } # add/overwrite
rem       = { type = "RRSIG" }        # remove if exists
```

Both could be combined to do something like this:
```toml
# Turn SERVFAIL into NXDOMAIN
[groups.nxdomain-write]
type      = "raw-write"
resolvers = [ "cloudflare" ]
add       = { status = "NXDOMAIN" }
rem       = { error = "EDE" }

[group.servfail-read]
type      = "raw-read"
resolvers = [ "cloudflare" ]
match     = { status = "SERVFAIL" }
action    = [ "nxdomain-write" ]
```

In this example you could also just use `static-responder` though.

folbricht:
Thanks for coming back to this with the follow-up, and for being straight that the workarounds covered what you actually needed.

I don't think `raw-read` and `raw-write` are the right shape for this. To be generally useful they would need a match language and a mutation language expressed in TOML, covering DNS header flags, EDNS flags, rcodes, record types and EDNS options, with each new field being another schema addition. At that point the config file has become a programming language with worse ergonomics than an actual one. That expressiveness is the reason the Lua group exists, and I would rather keep the TOML elements narrow and well defined and let Lua handle the open-ended cases.

[...]

PR #444, opened 2025-03-22
Failback immediately

Anuskuss:
I want to turn `SERVFAIL`s into `NXDOMAIN` so my client cache the response (you're not supposed to cache `SERVFAIL`s at all or for [no longer than 5 minutes](https://datatracker.ietf.org/doc/html/rfc2308#section-7.1)).

The easiest way to achive that (besides #443) would be to just `fail-back` to a `static-responder` with `rcode=3` and `ns=[". 3600 IN SOA . . 0 3600 3600 3600 3600"]`. But I don't want to keep using the `static-responder` and 1 second is too long for switching back, so this change allows you to use it for 0 seconds (i.e. only switch to the `static-resonder` once).

### DIFF
--- a/cmd/routedns/config.go
+++ b/cmd/routedns/config.go
@@ -109,7 +109,7 @@ type group struct {
 	EDNS0Data  []byte                  `toml:"edns0-data"`  // EDNS0 modifier option data
 
 	// Failover/Failback options
-	ResetAfter    int  `toml:"reset-after"`    // Time in seconds after which to reset resolvers in fail-back and random groups, default 0 (reset immediately).
+	ResetAfter    int  `toml:"reset-after"`    // Time in seconds after which to reset resolvers in fail-back and random groups, or -1 to reset immediately, default 60.
 	ServfailError bool `toml:"servfail-error"` // If true, SERVFAIL responses are considered errors and cause failover etc.
 	EmptyError    bool `toml:"empty-error"`    // If true, empty responses are considered errors and cause failover etc.
 
--- a/cmd/routedns/config.go
+++ b/cmd/routedns/config.go
@@ -111,7 +111,7 @@ type group struct {
 	// Failover/Failback options
 	ResetAfter    int  `toml:"reset-after"`    // Time in seconds after which to reset resolvers in fail-back and random groups, default 0 (reset immediately).
 	ServfailError bool `toml:"servfail-error"` // If true, SERVFAIL responses are considered errors and cause failover etc.
-	EmptyError    bool `toml:"empty-error"`    // If true, empty responses that are unusual (such as NOERROR with no records instead of NXDOMAIN) are considered errors and cause failover etc.
+	EmptyError    bool `toml:"empty-error"`    // If true, empty responses are considered errors and cause failover etc.
 
 	// Cache options
 	Backend                  *cacheBackend
--- a/cmd/routedns/config.go
+++ b/cmd/routedns/config.go
@@ -109,7 +109,7 @@ type group struct {
 	EDNS0Data  []byte                  `toml:"edns0-data"`  // EDNS0 modifier option data
 
 	// Failover/Failback options
-	ResetAfter    int  `toml:"reset-after"`    // Time in seconds after which to reset resolvers in fail-back and random groups, or -1 to reset immediately, default 60.
+	ResetAfter    int  `toml:"reset-after"`    // Non-zero time in seconds after which to reset resolvers in fail-back and random groups, or negative to reset immediately, default 60.
 	ServfailError bool `toml:"servfail-error"` // If true, SERVFAIL responses are considered errors and cause failover etc.
 	EmptyError    bool `toml:"empty-error"`    // If true, empty responses are considered errors and cause failover etc.
 
--- a/cmd/routedns/config.go
+++ b/cmd/routedns/config.go
@@ -111,6 +111,7 @@ type group struct {
 	// Failover/Failback options
 	ResetAfter    int  `toml:"reset-after"`    // Time in seconds after which to reset resolvers in fail-back and random groups, default 0 (reset immediately).
 	ServfailError bool `toml:"servfail-error"` // If true, SERVFAIL responses are considered errors and cause failover etc.
+	EmptyError    bool `toml:"empty-error"`    // If true, empty responses that are unusual (such as NOERROR with no records instead of NXDOMAIN) are considered errors and cause failover etc.
 
 	// Cache options
 	Backend                  *cacheBackend
--- a/cmd/routedns/config.go
+++ b/cmd/routedns/config.go
@@ -109,7 +109,7 @@ type group struct {
 	EDNS0Data  []byte                  `toml:"edns0-data"`  // EDNS0 modifier option data
 
 	// Failover/Failback options
-	ResetAfter    int  `toml:"reset-after"`    // Time in seconds after which to reset resolvers in fail-back and random groups, default 60.
+	ResetAfter    int  `toml:"reset-after"`    // Time in seconds after which to reset resolvers in fail-back and random groups, default 0 (reset immediately).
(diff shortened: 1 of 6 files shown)


folbricht:
There's a couple of issues with this change (both can be fixed).
1. It changes the default value which will impact any config relying on it. So we ideally shouldn't be changing it. How about using `-1` instead to mean "only fail over for this one request"? Please also include a mention of it in the [docs](https://github.com/folbricht/routedns/blob/master/doc/configuration.md#fail-back-group)
2. The current code will still fail-over and depending on timing some other requests may be going to the fallback as well which is probably not what you want. I think the easiest way to fix that is prevent it from changing the "active" resolver at all by wrapping https://github.com/folbricht/routedns/blob/master/failback.go#L95 into a condition like `if opt.ResetAfter == -1 {` or similar. That way, failing over never impacts subsequent requests.

Anuskuss:
1. I wasn't going to change the default behaviour but I'm not good at Go and I couldn't find the code that initializes the value to `0` in the first place. It would make more sense to do `if timeout == NULL; then timeout = 60` and if `0` is set explicitly then don't touch it at all. But I don't know how.
2. I tested my change with `while true; do dig +short bad; done` and `while true; do dig +short good; done` for a minute and the values were all good. But I guess a special case for `0` or `-1` couldn't hurt.

Anuskuss:
In my personal build I have also extended the `isSuccessResponse` to detect empty `NOERROR`s (gated behind `servfail-error=true`):

```diff
--- a/failback.go
+++ b/failback.go
@@ -158,5 +155,5 @@ func (r *FailBack) startResetTimer() chan struct{} {
 
 // Returns true is the response is considered successful given the options.
 func (r *FailBack) isSuccessResponse(a *dns.Msg) bool {
-       return a == nil || !(r.opt.ServfailError && a.Rcode == dns.RcodeServerFailure)
+       return a == nil || !(r.opt.ServfailError && (a.Rcode == dns.RcodeServerFailure || a.Rcode == dns.RcodeSuccess && len(a.Answer) == 0))
 }
```

It's probably not useful to you but I found that `curl` doesn't cache those responses (as opposed to `NXDOMAIN`).

Anuskuss:
I added the ability to treat "unusual" empty reponses (so far `NOERROR` with no records and `NXDOMAIN` with just a `CNAME` record) as errors. I've also updated the docs.
It seems like Go has no such concept as default values in structs (without jumping through hoops like labels) so I think I'll just go ahead and break the promise that nothing means a minute.

Anuskuss:
Actually nevermind, this is not enough for my use case. This now turns all empty responses into errors.

Anuskuss:
I found yet another case: Multiple `CNAME`s with no records. I don't want to loop through the whole list of answers because that sounds like pointless work (every milliseconds counts) so I'm just checking the first and last record, hoping that nobody is insane enough to return `CNAME` + `A` + `CNAME`.

failback.go
```diff
@@ -158,5 +159,8 @@ func (r *FailBack) startResetTimer() chan struct{} {
 
 // Returns true is the response is considered successful given the options.
 func (r *FailBack) isSuccessResponse(a *dns.Msg) bool {
-	return a == nil || !(r.opt.ServfailError && a.Rcode == dns.RcodeServerFailure)
+	aLen := len(a.Answer)
```
folbricht, line 162: If `a` is nil here, this will cause a nil-pointer panic. You can instead just add `len(a.Answer)` where you need it below. Or break out the nil-check first.

failback.go
```diff
@@ -158,5 +159,8 @@ func (r *FailBack) startResetTimer() chan struct{} {
...
+	return a == nil || !(r.opt.ServfailError && a.Rcode == dns.RcodeServerFailure) &&
```
folbricht, line 163: This is very hard to read and it looks wrong as well, should probably be `a == nil || !(r.opt.ServfailError && a.Rcode == dns.RcodeServerFailure) || ...`. It might be best to rewrite it with simple if statements and avoid the boolean algebra here:

```go
if a == nil {
  return true
}
if !(r.opt.ServfailError && a.Rcode == dns.RcodeServerFailure) {
  return true
}
if r.opt.EmptyError {
...
}
```

cmd/routedns/config.go
```diff
@@ -109,8 +109,9 @@ type group struct {
 	EDNS0Data  []byte                  `toml:"edns0-data"`  // EDNS0 modifier option data
 
 	// Failover/Failback options
-	ResetAfter    int  `toml:"reset-after"`    // Time in seconds after which to reset resolvers in fail-back and random groups, default 60.
+	ResetAfter    int  `toml:"reset-after"`    // Time in seconds after which to reset resolvers in fail-back and random groups, default 0 (reset immediately).
```
folbricht, line 112: This is changing the default. In Go, a value that isn't being assigned has a default value. In the case of an `int` that would be 0. That's why https://github.com/folbricht/routedns/blob/master/failback.go#L64-L66 is setting the default when the value hasn't been configured in a config. You can switch to using -1 though.

folbricht (changes requested):
One of my previous comment must have disappeared when a new change was pushed. This PR still needs a small change (probably early in https://github.com/folbricht/routedns/blob/master/failback.go#L115) to make sure it doesn't actually fail over to the next resolver. I suspect something as simple as this will do the trick.
```Go
if opt.ResetAfter == -1 {
  return
}
```

failback.go
Anuskuss: No, the logic was correct and I think the code is clever because there's no redundancy (the rest of the condition only runs when it hasn't hit bad case). The new code is double the lines for no real benefit other than that it's maybe a tiny bit faster to understand.

cmd/routedns/config.go
Anuskuss: If you insist on that then alright I guess. Now it accepts any negative integer. The text is wrong now though because somebody might still want to use 0 because they read "time in seconds" and think 0 will mean 0 seconds.

Anuskuss:
> This PR still needs a small change (probably early in https://github.com/folbricht/routedns/blob/master/failback.go#L115) to make sure it doesn't actually fail over to the next resolver. I suspect something as simple as this will do the trick.
> 
> ```go
> if opt.ResetAfter == -1 {
>   return
> }
> ```

I really have no idea where that check would make sense.  Admittedly I don't understand the code very well but if you start a timer with 0 seconds or less, it will resolve immediately. Or do you mean the case where it jumps from `resolver-1` to `resolver-2` to `resolver-3`?

cmd/routedns/config.go
folbricht: True, that could be clarified in the doc. Basically "not set" means 0, which will turn into 60s as a default. I just want to be very careful about breaking existing configs that do not set this value. It'd change the behavior quite a bit if 0 was actually used.

Anuskuss: Yeah well I would've preferred to set it to 60 only if it's undefined/not set. But no idea how to code that up in Go. If you could write the code for that that'd be great. Because now I have to tell the user to pick a positive number, but some users may think that this includes 0, so it's getting pretty convoluted.

folbricht:
It may be possible to distinguish between `0` and "unset". If so we could use an explicit value of `0` in the config to mean "immediate reset'. I can try that, but not today

Anuskuss:
I have no idea if that does what it's supposed to do.
First of all it's turning my `-1` into `-1000000000` so there's that. Also I tried to test it with `1` to make sure it even does something but that doesn't work. So I have no idea if the `-1` case works by accident or not.

folbricht:
I'll take a look at your PR and update it (possibly dealing with the unset vs 0 issue as well)

Anuskuss:
Just take the code. This is a joke username anyway so no need to credit me.

Anuskuss:
FWIW the old condition compiled to 27020349 bytes while the new one only uses 27016445. So it might actually be faster.

folbricht:
Moved to https://github.com/folbricht/routedns/pull/448

PR #453, opened 2025-04-20
Support for Lua scripting

folbricht:
WIP

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,6 +42,7 @@ require (
 	github.com/quic-go/qpack v0.5.1 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/txthinking/runnergroup v0.0.0-20230325130830-408dc5853f86 // indirect
+	github.com/yuin/gopher-lua v1.1.2-0.20241109074121-ccacf662c9d2 // indirect
 	go.uber.org/mock v0.4.0 // indirect
 	golang.org/x/crypto v0.35.0 // indirect
 	golang.org/x/exp v0.0.0-20240506185415-9bf2ced13842 // indirect
--- a/go.sum
+++ b/go.sum
@@ -92,6 +92,8 @@ github.com/txthinking/runnergroup v0.0.0-20230325130830-408dc5853f86/go.mod h1:c
 github.com/txthinking/socks5 v0.0.0-20230325130024-4230056ae301 h1:d/Wr/Vl/wiJHc3AHYbYs5I3PucJvRuw3SvbmlIRf+oM=
 github.com/txthinking/socks5 v0.0.0-20230325130024-4230056ae301/go.mod h1:ntmMHL/xPq1WLeKiw8p/eRATaae6PiVRNipHFJxI8PM=
 github.com/yuin/goldmark v1.4.13/go.mod h1:6yULJ656Px+3vBD8DxQVa3kxgyrAnzto9xy5taEt/CY=
+github.com/yuin/gopher-lua v1.1.2-0.20241109074121-ccacf662c9d2 h1:PF/PSu+RcSOzNpCqGo6KEO+4qw+LEiX/oqVxGm8rup8=
+github.com/yuin/gopher-lua v1.1.2-0.20241109074121-ccacf662c9d2/go.mod h1:GBR0iDaNXjAgGg9zfCvksxSRnQx76gclCIb7kdAd1Pw=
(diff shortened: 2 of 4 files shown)


Anuskuss:
I haven't updated in a long time but I just did and checked the changelog and saw this. And I gotta say: Thank you for your work, this is great :+1: 
I'll have to go back and reread all of my issues and check my past ideas. I think a bunch  fo them were already solved (by for example the introduction of templates) but two things that immediately came to mind were an `ANY` emulator (where with the help of Lua you'd be able to go through all possible RRs, query/collect them all and return) and a version check similar to [`version.bind`](https://kb.isc.org/docs/aa-00307) (just for fun).

The latter looks something like this:
```toml
[groups.version]
type           = "lua"
lua-script     = "function Resolve(m, c) a = Message.new():set_reply(m); a.answer = { RR.new({ rtype = TypeTXT, class = ClassCH, name = m.questions[1].name, txt = { string.match(io.popen('routedns -v'):read('*a'), 'Version:%s+(%S+)') } }) }; return a, nil; end"
lua-no-sandbox = true

[routers.router]
routes = [
  { resolver = "version", type = "TXT", class = "CH", name = '(?i)^version\.routedns\.$' }
]
```

folbricht:
@Anuskuss Glad it's useful. As for your questions:

### Version check

That should already work since you're using `lua-no-sandbox = true`, but https://github.com/folbricht/routedns/pull/494 improves it by directly exporting the version to Lua, so you can run this with the sandbox enabled.

```toml
groups.version]
type = "lua"
lua-script = """
function Resolve(msg, ci)
    local a = Message.new():set_reply(msg)
    a.answer = {
        RR.new({
            rtype = TypeTXT,
            class = ClassCH,
            name  = msg.questions[1].name,
            ttl   = 0,
            txt   = {BuildVersion},
        })
    }
    return a, nil
end
"""

[routers.router]
routes = [
  { resolver = "version", type = "TXT", class = "CH", name = '(?i)^version\.routedns\.$' },
  { resolver = "cloudflare-dot" },
]
```

### ANY emulator

I think this too is already possible. You could iterate over a list of record types, query each one, and merge the answers. 

```toml
[groups.any-emulator]
type       = "lua"
resolvers  = ["upstream"]
lua-script = """
function Resolve(m, c)
  local types = {TypeA, TypeAAAA, TypeMX, TypeTXT, TypeCNAME, TypeNS, TypeSOA, TypeSRV}
  local reply = Message.new():set_reply(m)
  reply.answer = {}
  for _, t in ipairs(types) do
    local q = Message.new()
    q.questions = { Question.new(m.questions[1].name, t) }
    q.recursion_desired = true
    local r, err = Resolve(q, c, resolvers[1])
    if r and #r.answer > 0 then
      for _, rr in ipairs(r.answer) do
        reply.answer[#reply.answer + 1] = rr
      end
    end
  end
  return reply, nil
end
"""
```

This is somewhat expensive (one upstream query per type) but should work.

Anuskuss:
> You have to use `:` for methods in Lua

Ah dang, that's embarrassing :P Even though I myself used `:` already. I guess my brain just saw the `.` and though: Yep, that's how you call a method.

> For the sub-queries, copy `rd` with `q.recursion_desired = msg.recursion_desired`

That makes sense.

> it should probably set `ra` to true

Can you make that the default then? Meaning `:resolve()` should do exactly the same thing that RouteDNS would've done (even if wrong). Just so regular users doesn't have to think about what's right or wrong.
But that doesn't answer the full question though: Should a DNS server return `ra` when `rd` wasn't sent? And what sense does it make to send `ra` when we didn't even get a `ra` response from upstream? Unless that bit is to be taken at face value, meaning: "we support recursion".

Regarding question 4: I personally don't care for anything other than `A` and `AAAA`. I might make an exception of `TXT` though (I use that for one of my domains). But still, I'm only getting a single `A` record. Maybe that's enough  to fulfill the `ANY` "promise"? Maybe I'll reuse the `ANY` code and introduce an `ALL` RR, which doesn't stop querying until a record repeats (but still only for `A` and `AAAA`).

I'm also wondering if the `err == nil` check is superfluous. Errors are irrelevant here because we are trying to collect *any* responses possible. And since in the case of an error `ipairs(a.answer)` just evaluates to `{}` it probably won't cost much (and you save a comparison). The `a ~= nil` looks intriguing though: Can that even happen or are you just being extra cautious?

PR #472, opened 2025-11-10
Add async Redis SET on cache miss for reduced latency

emlimap:
Implement async Redis SET on cache miss to reduce miss-path latency by removing the Redis write from the critical request path.

  ## Changes
  - **Config flag**: `redis-async-set-on-miss` (default: `true`)
  - **Behavior**: On cache miss, return upstream response immediately and perform Redis SET in background
  - **Concurrency control**: Semaphore (256 slots) caps concurrent async writes
  - **Metrics**: `async-skipped` counter tracks when semaphore is saturated
  - **TTL guard**: Skip storing already-expired items in Redis

  ## Benefits
  - Reduces miss-path latency by ~100ms (typical Redis SET time)
  - Bounded goroutine usage prevents resource amplification
  - Memory backend and cache hits unchanged
  - Backward compatible with opt-out via config

  ## Trade-offs
  Brief window where duplicate queries may both miss cache until background write completes.

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -13,11 +13,6 @@ import (
 	"github.com/miekg/dns"
 )
 
-const (
-	// asyncWriteSemCapacity limits concurrent background Redis writes on cache miss.
-	asyncWriteSemCapacity = 256
-)
-
 // Cache stores results received from its upstream resolver for
 // up to TTL seconds in memory.
 type Cache struct {
@@ -26,8 +21,6 @@ type Cache struct {
 	resolver Resolver
 	metrics  *CacheMetrics
 	backend  CacheBackend
-	// Semaphore to limit concurrent async Redis writes (buffered channel).
-	asyncWriteSem chan struct{}
 }
 
 type CacheMetrics struct {
@@ -37,8 +30,6 @@ type CacheMetrics struct {
 	miss *expvar.Int
 	// Current cache entry count.
 	entries *expvar.Int
-	// Count of async stores skipped due to semaphore saturation.
-	asyncSkipped *expvar.Int
 }
 
 var _ Resolver = &Cache{}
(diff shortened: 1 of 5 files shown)


cache.go
```diff
@@ -208,7 +219,27 @@ func (r *Cache) Resolve(q *dns.Msg, ci ClientInfo) (*dns.Msg, error) {
 
 	// Put the upstream response into the cache and return it. Need to store
 	// a copy since other elements might modify the response, like the replacer.
-	r.storeInCache(q, a.Copy())
+	// For Redis backends with async-set-on-miss enabled, store asynchronously
+	// to reduce miss-path latency.
+	if rb, ok := r.backend.(*redisBackend); ok && rb.opt.AsyncSetOnMiss {
```
folbricht: Not a fan of doing the type cast here. Ideally, this code should go into cache-redis.go (in `Store()`). Then there's no casting necessary and the cache doesn't need to know/care about what kind of backend is being used.

cmd/routedns/config.go
```diff
@@ -92,6 +92,7 @@ type cacheBackend struct {
 	RedisMaxRetries      int    `toml:"redis-max-retries"`       // Maximum number of retries before giving up. Default is 3 retries; -1 (not 0) disables retries.
 	RedisMinRetryBackoff int    `toml:"redis-min-retry-backoff"` // Minimum back-off between each retry. Default is 8 milliseconds; -1 disables back-off.
 	RedisMaxRetryBackoff int    `toml:"redis-max-retry-backoff"` // Maximum back-off between each retry. Default is 512 milliseconds; -1 disables back-off.
+	RedisAsyncSetOnMiss  *bool  `toml:"redis-async-set-on-miss"` // When true (default), perform Redis SET asynchronously on cache miss to reduce latency.
```
folbricht: I think we should invert this to have a default of false which is more aligned with the rest, something where false == write asynchronously. Also, the `OnMiss` in the name is confusing. Maybe just `RedisSyncSet`?

folbricht (commented):
This should work, but I'd like to move the async logic into the cache (this will clash with https://github.com/folbricht/routedns/pull/473, so would need to be done sequentially). Also the name+default of the config option should be changed.

I can make the if you like. Are you able to test this as well? I'm not currently setup with all pieces needed.

PR #473, opened 2025-11-10
Redis cache: use binary wire format with pooled buffers

emlimap:
Store Redis cache entries in compact binary wire format to reduce payload size and CPU usage. Uses `dns.Msg.PackBuffer` with adaptive `sync.Pool` to minimize allocations.

  ## Changes
  - Binary format: 10-byte header (version, flags, timestamp) + DNS wire bytes
  - Adaptive buffer pooling (2 KiB → 32 KiB cap) prevents repeated allocations
  - Serialization moved outside I/O deadline for accurate network timeouts
  - Backward compatible: auto-detects and decodes legacy JSON entries

  ## Benefits
  - Smaller Redis storage footprint
  - Lower CPU on cache hits/misses
  - Reduced allocations via pooled buffers
  - Safe mixed deployments (old JSON + new binary)

### DIFF
--- a/cache-redis.go
+++ b/cache-redis.go
@@ -64,21 +64,23 @@ func encodeCacheAnswer(item *cacheAnswer) ([]byte, error) {
 	}()
 
 	if cap(buf) == 0 {
-		buf = make([]byte, 2048)
+		buf = make([]byte, 0, 2048)
 	}
 
-	// Use the full capacity of the buffer for PackBuffer so it can reuse it.
+	// Pack DNS message first into the scratch buffer
 	buf = buf[:cap(buf)]
-	packed, err := item.Msg.PackBuffer(buf)
+	dnsWire, err := item.Msg.PackBuffer(buf)
 	if err != nil {
 		return nil, fmt.Errorf("failed to pack DNS message: %w", err)
 	}
 
-	// Keep the (potentially grown) buffer in the pool for future calls.
-	buf = packed
+	// Keep the (potentially grown) buffer for cleanup
+	buf = dnsWire
 
-	// Allocate result buffer: header (10 bytes) + packed DNS message.
-	result := make([]byte, headerSize+len(packed))
+	// Allocate result with header + DNS wire bytes
+	result := make([]byte, headerSize+len(dnsWire))
+
+	// Write header
 	result[0] = binaryFormatVersion
 
 	var flags byte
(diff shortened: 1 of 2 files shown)


cache-redis.go
```diff
@@ -50,18 +152,35 @@ func (b *redisBackend) Lookup(q *dns.Msg) (*dns.Msg, bool, bool) {
...
+				Log.Error("failed to decode cache record from redis (binary and JSON)", "binary_error", err, "json_error", jsonErr)
+				return nil, false, false
+			}
+		} else {
+			// Try JSON anyway
+			if jsonErr := json.Unmarshal(valueBytes, &a); jsonErr != nil {
+				Log.Error("failed to decode cache record from redis", "error", err)
+				return nil, false, false
+			}
+		}
```
folbricht: The conditional here looks unnecessary, it'll unmarshal with JSON regardless, if it starts with `{` or not.

cache-redis.go
```diff
@@ -24,6 +26,103 @@ type RedisBackendOptions struct {
...
+	New: func() any {
+		return make([]byte, 0, 2048)
+	},
+}
+
+const (
+	binaryFormatVersion = 1
+	headerSize          = 10
+	flagPrefetchBit     = 1 << 0
+	maxPooledBufCap     = 32 << 10 // 32 KiB - cap pooled buffers to prevent unbounded growth
```
folbricht: Not sure the limiter is necessary

cache-redis.go
```diff
@@ -24,6 +26,103 @@ type RedisBackendOptions struct {
...
+// - bytes 10..N: dns.Msg wire bytes
+func encodeCacheAnswer(item *cacheAnswer) ([]byte, error) {
+	// Get a buffer from the pool for packing the DNS message
+	buf := packBufPool.Get().([]byte)
+
+	// Pack the DNS message - PackBuffer may return a larger slice if needed
+	packed, err := item.Msg.PackBuffer(buf)
+	if err != nil {
+		// Return buffer to pool even on error
+		packBufPool.Put(buf[:0])
```
folbricht: Using `defer` is more common, no need to cover every code-path that returns

folbricht (commented):
There are a few inconsistencies and some unnecessary complexity in this. Since you mentioned you're not comfortable writing Go, would  you prefer I make a PR against yours, or just take this code and fix it up?

emlimap:
@folbricht The failed to decode messages was because the format being written to Redis was corrupted. This is now fixed. Got Claude code to take your PR comments into account when making the changes.

It also added tests for encoding and decoding. Not sure if it is useful for you. If not, feel free to remove it.

PR #471, opened 2025-10-20
Add ECS source support to routing (#470)

tekert:
Fixes #470

Tested the build on my router and works perfectly.

Created a new test don't know if that's correct or you want it integrated on source test.

PR #476, opened 2025-12-09
Add ECS Source Routing documentation to configuration guide

tekert:
Draft for missing documentation on #471 merge


tekert:
The examples in the documentation use /32 masks because they represent individual IPs, which are the most common use case for per-client policies. However, the feature is not limited to /32. You can use any valid CIDR notation to define subnets.

```go
[routers.ecs-router]
routes = [
  # Match all clients in the 192.168.1.0/24 subnet
  { ecs-source = "192.168.1.0/24", resolver = "subnet-resolver" },

  # Match a specific client IP
  { ecs-source = "192.168.1.10/32", resolver = "specific-client-resolver" },

  # Default route for all other queries
  { resolver = "default-resolver" },
]
```

PR #484, opened 2026-01-12
Add configurable idle-timeout for DoH connections

emlimap:
This PR adds support for configuring the DoH idle timeout, which wasn't previously configurable. I found it helpful for low-volume query scenarios—keeping an existing connection open for 60 seconds performed slightly better than letting it expire at the default 30 seconds and opening a new one.

* TCP transport: defaults to 30 seconds
* QUIC transport: uses quic-go library default unless explicitly configured (actual timeout is min of client/server values)

PS: Feel free to make changes to this PR as you see fit should you need to.

### DIFF
--- a/cmd/routedns/config.go
+++ b/cmd/routedns/config.go
@@ -73,7 +73,8 @@ type resolver struct {
 
 // DoH-specific resolver options
 type doh struct {
-	Method string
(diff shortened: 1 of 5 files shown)


folbricht (approved):
Thank you

PR #500, opened 2026-03-11
Add tests for isSuccessResponse()

folbricht:
## Summary

- Adds unit tests for the `isSuccessResponse()` method on `FailBack`, covering all branches of the refactored logic from #497
- Tests cover: nil responses, SERVFAIL/REFUSED/NOTIMP handling, EmptyError option, answer type matching (including CNAME chains), EDE blocked/censored/filtered codes, and TypeANY with HINFO

Depends on #497.

### DIFF
--- a/failback_test.go
+++ b/failback_test.go
@@ -297,11 +297,33 @@ func TestFailBackIsSuccessResponse(t *testing.T) {
 		require.False(t, fb.isSuccessResponse(msg))
 	})
 
+	t.Run("EmptyAnswerWithEDEProhibited", func(t *testing.T) {
+		fb := newFB(FailBackOptions{EmptyError: true})
+		msg := newMsg(dns.TypeA, dns.RcodeSuccess)
+		addEDE(msg, dns.ExtendedErrorCodeProhibited)
+		require.False(t, fb.isSuccessResponse(msg))
+	})
+
+	t.Run("EmptyAnswerWithInvalidEDE", func(t *testing.T) {
+		fb := newFB(FailBackOptions{EmptyError: true})
+		msg := newMsg(dns.TypeA, dns.RcodeSuccess)
+		addEDE(msg, 9999)
+		require.False(t, fb.isSuccessResponse(msg))
+	})
+
 	t.Run("TypeANYWithHINFO", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


Anuskuss:
Yeah but that's not what's supposed to happen: In the default configuration (`EmptyError=false`) an `NXDOMAIN` will be successful, therby not causing a rotation. A rotation in this case doesn't make sense. If the user queries a nonexistent domain, the first server will obviously return `NXDOMAIN`. Now, because of the new logic, a rotation will happen. The second server will also obviously return `NXDOMAIN`. Another rotation. You see how this is just wasting CPU and bandwidth, right?
The exception is when `EmptyError=true` is true. In this case you don't care about the wasted resources, you just want an answer. But there is a reason why this isn't the defaut case, as it's only useful in specific scenarios.

Anuskuss:
> So basically that means:
> 
> ```
> NXDOMAIN + EmptyError=true + no EDE => rotate
> 
> NXDOMAIN + EmptyError=true + EDE => no rotation
> ```
Yes. EDE (15-17) gives an empty response meaning ("this reponse is empty because we blocked it"). `NXDOMAIN` is the normal response for when the query can't *actually* be fulfilled (these days nobody blocks a domain without *either* EDE or `0.0.0.0`) and should be treated the same as empty responses with `NOERROR`. `EmptyError` just says that "we don't care what you say, we want a non-empty reponse".

> Isn't that a bit too sensitive? NXDOMAIN responses are quite common in normal use, and if `EmptyError=true` that would rotate every time.

`EmptyError` shouldn't be used (and why it's disabled by default) unless in specific scenarios. I, for example, use it to override empty reponses with `0.0.0.0`. But in the general case, where you have e.g. `resolvers = [ "cloudflare", "google", "quad9" ]` it doesn't make sense to send the same query to all of them, as all of them will return `NXDOMAIN`. In the general case, you'd only want `failrotate`/`failback` because of **fail**ure, e.g. if `cloudflare` goes down (times out or `SERVFAIL`s), you'd want RouteDNS to rotate to `google`. And if `cloudflare` behaves correctly, but returns `NXDOMAIN`, that probably just means that the domain doesn't exist for real.

Issue #501, opened 2026-03-12
[Suggestion] Replace "local-address" in resolvers by local-address-v4 and local-address-v6


przemyslaw0:
`local-address` in a resolver section is used to choose the outgoing address for sending the DNS queries, but what if a named DNS resolver (like https://cloudflare-dns.com/dns-query) has both IPv4 and IPv6 addresses?

<img width="1019" height="368" alt="Image" src="https://github.com/user-attachments/assets/8d35ee94-0fd3-45e7-a20b-42be9ead4be2" />

This option can be specified only one time, but will cover only one IP type, IPv4 or IPv6. This could lead to a leakage case an IP of different type specified in `local-address` is queried by RouteDNS.

przemyslaw0:
**local-address-v4 = IPv4 address
local-address-v6 = IPv6 address**

PR #504, opened 2026-03-13
Add fwmark (SO_MARK) and bind-if (SO_BINDTODEVICE) support

### LINKED ISSUE #461: support firewall mark (fwmark) based routing for resolvers
So in a multi wan setup, it would be good to route certain resolvers one path and others in another. At the moment, the way this can be done is by running one or more socks5 proxy server that can bind to specific wan interfaces.

While this works, this is less efficient considering the socks5 overhead. Instead, it would be far more efficient to mark the outgoing packets with a certain mark and IPtables & IP route will take care of routing it across the correct interface.

At the moment I use sing-box with routing mark (https://sing-box.sagernet.org/configuration/shared/dial/#routing_mark) which listens as socks5 proxy and it would be good to remove it from the loop. It is also written in Go so may provide you with some pointers if needed https://github.com/SagerNet/sing-box

folbricht:
## Summary

- Adds `fwmark` and `bind-if` config options to all resolvers and listeners, enabling Linux policy-based routing and VRF support without an external SOCKS5 proxy
- New `SocketOptions` struct with platform-specific implementations (`sockopts_linux.go` / `sockopts_other.go`) using `SO_MARK` and `SO_BINDTODEVICE`
- For client `net.Dialer` connections, options are applied via the `Control` callback; for UDP and listener sockets, options are applied post-creation via `SyscallConn`
- DTLS listeners log a warning since the pion/dtls library creates sockets internally

Closes #461

### Configuration example

```toml
[resolvers.quad9-wan2]
address = "9.9.9.9:53"
protocol = "udp"
fwmark = 11
bind-if = "eth2"

[listeners.vrf-udp]
address = ":53"
protocol = "udp"
resolver = "quad9-wan2"
bind-if = "vrf-blue"
```

### DIFF
--- a/cmd/routedns/config.go
+++ b/cmd/routedns/config.go
@@ -36,7 +36,7 @@ type listener struct {
 	OdohMode   string   `toml:"odoh-mode"` // ODoH mode - accepts "proxy", "target" or "dual", default is target mode
 	AllowDoH      bool     `toml:"allow-doh"` // Allow ODoH listeners to also handle DoH queries to /dns-query
 	NetNS         string   `toml:"netns"`     // Linux network namespace name or absolute path
-	FWMark        int      `toml:"fwmark"`    // Linux firewall mark (SO_MARK) for the listening socket
+	FWMark        uint32   `toml:"fwmark"`    // Linux firewall mark (SO_MARK) for the listening socket
 	BindInterface string   `toml:"bind-if"`   // Linux network interface to bind the socket to (SO_BINDTODEVICE)
 	Frontend      dohFrontend
 }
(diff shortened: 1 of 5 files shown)


folbricht:
Thank you. The control options needed to be applied before bind(). Can you try again? Hope this didn't break anything else.

przemyslaw0:
100% tested and everything work as expected, I believe it's ready to merge.

The next minor fix would be this: https://github.com/folbricht/routedns/issues/501

It can lead to leaks in resolvers (packet leaks to wrong network interfaces).

PR #517, opened 2026-04-17
Add domain-subdomain blocklist format (#515)

### LINKED ISSUE #515: Feature Request: Alternative domain blocklist format or an include sub-domains option
I like using [hagezi's blocklists](https://github.com/hagezi/dns-blocklists). Of the formats provided, there are two that work mostly-well for RouteDNS:
* Wildcard Asterisk (contains entries like `*.example.com`)
* Wildcard Domains (contains entries like `example.com`)

Currently, in order to block `subdomain.example.com` and `example.com`, I would need to add *both* of these lists (which massively increases the memory usage). It'd be nice to have an alternative blocklist format or an option that would allow for only one of these lists to be used.

Thank you!

folbricht:
## Summary

Adds a new `domain-subdomain` blocklist format that treats bare entries (e.g. `example.com`) as matching the apex *and* all sub-domains — equivalent to a `.example.com` rule in the existing `domain` format. Entries that already use `*.` retain their sub-domain-only semantics and can be used as a per-entry opt-out.

This addresses #515: blocklists such as hagezi's "Wildcard Domains" lists, where every line is intended to block apex + sub-domains, can now be loaded as-is without rewriting each line or doubling memory by also loading the matching "Wildcard Asterisk" list.

## Behavior

| Entry            | `domain` format       | `domain-subdomain` format |
| ---------------- | --------------------- | ------------------------- |
| `example.com`    | apex only             | apex + sub-domains        |
| `.example.com`   | apex + sub-domains    | apex + sub-domains        |
| `*.example.com`  | sub-domains only      | sub-domains only          |

## Changes

- `blocklistdb-domain.go` — refactor rule ingestion into a private helper, add `NewDomainSubdomainDB`, preserve the flag on `Reload()`.
- `cmd/routedns/main.go` — wire `"domain-subdomain"` into the format dispatcher used by blocklist, allowlist, and `response-blocklist-name`.
- `blocklistdb-domain_test.go` — `TestDomainSubdomainDB` and `TestDomainSubdomainDBError`.
- `cmd/routedns/example-config/blocklist-domain-subdomain.toml` — new example.
- `doc/configuration.md` — describe the new format, extend the `blocklist-format` / `allowlist-format` enumerations, add an example.

## Backward compatibility

Zero impact on existing configs. The `domain` format and any config using it behave exactly as before; users opt in by changing one string value.

Closes #515

### DIFF
--- a/blocklistdb-domain.go
+++ b/blocklistdb-domain.go
@@ -13,18 +13,35 @@ import (
 // domain.com: matches just domain.com and not subdomains
 // .domain.com: matches domain.com and all subdomains
 // *.domain.com: matches all subdomains but not domain.com
+//
+// When includeSubdomains is true, bare entries like "domain.com" are treated as
+// ".domain.com" (apex and all sub-domains). Entries that already use a leading
+// "." or "*." prefix keep their original semantics. This is intended for
+// blocklists where every line is meant to block both the apex and sub-domains
+// (e.g. hagezi's "Wildcard Domains" lists).
 type DomainDB struct {
-	name   string
-	root   node
-	loader BlocklistLoader
+	name              string
+	root              node
+	loader            BlocklistLoader
+	includeSubdomains bool
 }
 
 type node map[string]node
 
 var _ BlocklistDB = &DomainDB{}
 
-// NewDomainDB returns a new instance of a matcher for a list of regular expressions.
+// NewDomainDB returns a new instance of a matcher for a list of domain rules.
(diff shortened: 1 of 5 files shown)


remilia-dev:
I was able to test this today & it works! Thank you!

I tested it with:
* the blocklists I normally utilize
* `blocklist` individual entries
* `allowlist` individual entries

With the `blocklist` & `allowlist` tests I also tested `example.com` vs `*.example.com` to ensure the behavior matches what you've documented. It all looks good to me!

folbricht:
Thanks for testing it.

PR #518, opened 2026-04-20
Save memory in domain blocklists by re-using specific nodes

remilia-dev:
Saves memory in domain & domain-subdomain blocklists by re-using nodes for empty ending nodes and second-to-last wildcard nodes (both `.` & `*.`).

This is especially significant for large lists. A minimal routedns config with just the [newly registered domains](https://github.com/hagezi/dns-blocklists#nrd) domain-subdomain list goes from using ~1.1GiB to ~300MiB with this change (resting memory usage after list loading; `GOMEMLIMIT` was used to encourage as much garbage collection as possible).

While the change-size is fairly small & the tests pass, this is my first time working with Go, so some adjustments might be necessary.

### DIFF
--- a/blocklistdb-domain.go
+++ b/blocklistdb-domain.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"maps"
 	"net"
-	"reflect"
-	"slices"
 	"strings"
 
 	"github.com/miekg/dns"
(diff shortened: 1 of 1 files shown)


folbricht:
Thank you. I'm not a fan of using "reflect" and unsafe pointers in this
case. It's quite heavy-handed (in Go, this should only be used when
absolutely necessary, like when operating on a low level) and may not be
necessary. It'd be better to implement this without. I can try to
re-implement this next weekend, unless you beat me to it.

folbricht:
https://github.com/remilia-dev/routedns/pull/1 is open against your PR, could you take a look?

remilia-dev:
I've merged your PR after testing it. Thank you! I'm glad I let you implement it without reflect as the way I was thinking of doing that was more involved/invasive than your simple change.

folbricht:
Thank you

Issue #516, opened 2026-04-14
Network namespaces


Anuskuss:
First of all thank you for the network namespaces support. I've been using it for a few weeks now and although it looks better than what I used to do before, I realized after rebooting that my setup is flawed. I won't go into detail about how my setup works or what I'm trying to achieve but it's a dependency issue:

My RouteDNS starts with the system and contains a resolver which should only exist in the namespace. Then there's my (systemd) service which creates my namespace (and runs my applications). The issue is that RouteDNS starts *before* the namespace is created, which results in errors. But I also can't run RouteDNS *after* the namespace because I need DNS as part of the setup routine.

Possible solutions:
1. Don't used `netns`. Through the use of a `source=""` `router`, an `ecs-op="add"` `group` (and an `iptables` rule) the same can be accomplished (this is what I used to do before). Although it's kinda ugly and requires a `router` which delays (even if it's just a tiny bit) *all* requests. 
2. Use a generated config with and without the namespace settings and change the config based on if the namespace already exists or not. The namespace service needs to restart / reload the RouteDNS service in that case.
3. Run 2 RouteDNS instances, one on bare metal and one in the namespace (without using `netns`). Systemd lets you run your services inside the namespace quite easily. Seems like overkill and a waste of CPU/RAM though.
4. RouteDNS should log the error but then instead of retrying once per second, it should `inotifywait` `/var/run/netns/<name>` and create the resolvers/listeners when the namespace becomes available. And then also destroy them when the namespace disappears. This would be my preferred solution but would require additional work from you.

My main issue is that I don't want RouteDNS to constantly keep retrying the namespace. Not only does it generate an error each second (filling my journal with `routedns[#]: time=# level=ERROR msg="listener failed" error="failed to open target netns \"#\": open /var/run/netns/#`) but it's also a waste of CPU cycles. There's also times where my namespace is not active at all so RouteDNS would literally just waste electricity. I also don't want to remember to restart / change the config of RouteDNS everytime I create my namespace. `inotifywait` should be basically free but I don't know if RouteDNS is currently capable of opening/closing ports and creating/destroying resolvers on-the-fly.

P.S. Oh and don't forget to add [CAP_SYS_ADMIN](https://github.com/folbricht/routedns/pull/499#issuecomment-4129435834) to https://github.com/folbricht/routedns/blob/3cde2ad13ffec69933e778df32c0fb2e78541b63/cmd/routedns/routedns.service#L10.

folbricht:
[...]

PR #522, opened 2026-04-26
Lazy-start listeners when network namespace is missing

folbricht:
## Summary
Closes the journal-spam complaint in #516. A listener configured with `netns = "<name>"` no longer retries `Start()` once per second when the namespace is absent. Instead, it warns once, watches `/var/run/netns/` via inotify, and starts the listener as soon as the namespace appears. If the namespace is later removed the listener is stopped cleanly; when it returns the listener is rebuilt and started again.

## Approach
- `Listener` gains `Stop() error`. Six of seven implementations already had one; only `DNSListener` needed adding (one-liner wrapping `dns.Server.Shutdown()`).
- New internal `netnsWatcher` (`netns_watch_linux.go`) — process-wide inotify watch on `/var/run/netns/`, demuxed per namespace name. Exported `SubscribeNetNS` / `NetNSState` API. Stub on non-Linux returns "not supported". No new module deps; uses `golang.org/x/sys/unix` already in `go.mod`.
- The startup loop in `cmd/routedns/main.go` is split: each `case` of the protocol switch produces a `build` closure that constructs a fresh listener instance. Listeners without `netns` are built once and run through the existing retry loop. Listeners with `netns` go through a new `superviseNetNSListener` helper that drives the lifecycle off `SubscribeNetNS` events.
- A fresh instance is built on each namespace-present event, because some listener types (DoH/Admin/ODoH wrap `http.Server`) cannot be restarted after `Shutdown` — `ListenAndServe` returns `ErrServerClosed` permanently.
- Resolvers/clients with `netns` are out of scope: they fail at dial time and don't cause spam.

## Behavior
- `netns "foo"` configured, namespace absent at startup → single `WARN netns "foo" not present, waiting`. No further log activity until state changes.
- `ip netns add foo` → `INFO netns present, starting listener` and the listener binds.
- `ip netns delete foo` → `WARN netns "foo" gone, stopping listener`, listener stopped cleanly.
- Re-create the namespace → listener rebuilt and started automatically.
- Multiple listeners sharing one namespace share a single inotify watch.

## Notes
- Listeners without `netns` are unaffected.
- Non-Linux build paths return the existing "namespaces only supported on Linux" error.

### DIFF
--- a/adminlistener.go
+++ b/adminlistener.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/quic-go/quic-go"
@@ -18,6 +19,7 @@ const adminServerTimeout = 10 * time.Second
 
 // AdminListener is a DNS listener/server for admin services.
 type AdminListener struct {
+	mu         sync.Mutex
 	httpServer *http.Server
 	quicServer *http3.Server
 
@@ -76,20 +78,23 @@ func (s *AdminListener) Start() error {
 
 // Start the admin server with TCP transport.
 func (s *AdminListener) startTCP() error {
-	s.httpServer = &http.Server{
+	httpServer := &http.Server{
 		Addr:         s.addr,
 		TLSConfig:    s.opt.TLSConfig,
 		Handler:      s.mux,
 		ReadTimeout:  adminServerTimeout,
 		WriteTimeout: adminServerTimeout,
 	}
+	s.mu.Lock()
+	s.httpServer = httpServer
+	s.mu.Unlock()
 
 	ln, err := ListenInNetNS(context.Background(), s.opt.NetNS, "tcp", s.addr, s.opt.SocketOptions)
 	if err != nil {
 		return err
 	}
 	defer ln.Close()
-	return s.httpServer.ServeTLS(ln, "", "")
+	return httpServer.ServeTLS(ln, "", "")
 }
 
 // Start the admin server with QUIC transport.
@@ -109,12 +114,15 @@ func (s *AdminListener) startQUIC() error {
 		udpConn.Close()
 		return err
 	}
-	s.quicServer = &http3.Server{
+	quicServer := &http3.Server{
 		TLSConfig:  s.opt.TLSConfig,
 		Handler:    s.mux,
 		QUICConfig: &quic.Config{},
 	}
-	return s.quicServer.ServeListener(quicLn)
+	s.mu.Lock()
+	s.quicServer = quicServer
+	s.mu.Unlock()
+	return quicServer.ServeListener(quicLn)
 }
 
 // Stop the server.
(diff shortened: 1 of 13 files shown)


Anuskuss:
Everything appears to be working (single warning, start-stop-start, start before ns). Thanks for implementing this :+1:

Anuskuss:
After rebooting I discovered that it's quite time-sensitive and that RouteDNS tried to bind the namespace too early. If I run my unmodified "namespace creator" service it fails every time now (I didn't change anything) but if I instead add a sleep after namespace creation but before DNS access (to give RouteDNS time to retry) it works:

```
Jun 01 01:19:08 systemd[1]: Started routedns.service - RouteDNS - DNS stub resolver and router.
Jun 01 01:19:08 routedns[9794]: time=2026-06-01T01:19:08.066+02:00 level=INFO msg="starting listener" id=udp protocol=udp4 addr=:5300
Jun 01 01:19:08 routedns[9794]: time=2026-06-01T01:19:08.067+02:00 level=INFO msg="starting listener" id=tcp protocol=tcp4 addr=:5300
Jun 01 01:19:08 routedns[9794]: time=2026-06-01T01:19:08.067+02:00 level=WARN msg="netns not present, waiting" netns=test id=udp_test
Jun 01 01:19:12 routedns[9794]: time=2026-06-01T01:19:12.979+02:00 level=INFO msg="netns present, starting listener" netns=test id=udp_test
Jun 01 01:19:12 routedns[9794]: time=2026-06-01T01:19:12.979+02:00 level=INFO msg="starting listener" id=udp_test protocol=udp4 addr=:53
Jun 01 01:19:12 routedns[9794]: time=2026-06-01T01:19:12.979+02:00 level=WARN msg="listener exited" id=udp_test error="failed to open target netns \"test\": open /var/run/netns/test: permission denied"
Jun 01 01:19:13 routedns[9794]: time=2026-06-01T01:19:13.979+02:00 level=INFO msg="retrying listener start" netns=test id=udp_test
Jun 01 01:19:13 routedns[9794]: time=2026-06-01T01:19:13.980+02:00 level=INFO msg="starting listener" id=udp_test protocol=udp4 addr=:53
```

This is the sequence of `inotifywait`:
```
/run/netns/ OPEN,ISDIR 
/run/netns/ CLOSE_NOWRITE,CLOSE,ISDIR 
/run/netns/ CREATE test
/run/netns/ OPEN test
/run/netns/ CLOSE_NOWRITE,CLOSE test
```

I tried to watch for `CLOSE` instead of `CREATE` (in `netns_watch_linux.go`) but that didn't help. Do you care to investigate this? Adding an arbitrary delay before the bind seems ugly and risky because if the delay is too big, my service might have already tried the DNS server, failed and subsequently deleted the namespace. However, adding 50 ms seems to work (until it doesn't). I'm not sure there's a good solution for this.

folbricht:
@Anuskuss Would you mind trying out https://github.com/folbricht/routedns/pull/562?

Issue #537, opened 2026-05-12
Support health monitoring in resolver load balancing


emlimap:
Currently, when you have a group of resolvers, your options are limited to round robin, random, fail-rotate, fail-back, and fastest. While these all work, there's a problem when one or two resolvers experience latency and/or timeout issues that aren't currently accounted for. Recently, one of my resolvers timed out, and any queries routed to that resolver failed, causing issues for end devices.

In AdGuard Home, it offers a load balancing option that factors in response times and failures. Here are all of their options:

<img width="1150" height="253" alt="Image" src="https://github.com/user-attachments/assets/43243141-b234-4a1b-84ec-c2df87f623f3" />

It is also an open source project written in Golang https://github.com/AdguardTeam/AdGuardHome/

The DNS part like the above options are handled by this https://github.com/AdguardTeam/dnsproxy

I also think renaming `fastest` to `parallel` would be more accurate. However, that would be a breaking change for users currently using that option.

folbricht:
Added as a new `load-balance` group in #538. It tracks a per-resolver exponential moving average of response time and selects by weighted random, so slower resolvers get proportionally less traffic.

Failures are handled by `failure-penalty`, which records a configurable RTT penalty for a resolver that fails repeatedly, applied only after two consecutive failures so a single slow response does not suppress it. A fixed fraction of picks is made uniformly at random so penalized resolvers keep being re-measured and can recover rather than being starved.

See `example-config/load-balance.toml` and the docs in `doc/groups.md`.

Closing as done.

PR #538, opened 2026-05-15
add load-balance group with weighted random resolver selection

emlimap:
Had a go at https://github.com/folbricht/routedns/issues/537 with the help of AI. Please feel free to make changes if needed.

# Add load-balance group

Adds a new `load-balance` group type that routes queries using weighted random
selection, preferring resolvers with lower average response times. Failed
resolvers are penalized and the request is retried with the remaining resolvers,
similar to `fail-rotate` but without a fixed order.

## How it works

Each resolver is assigned a weight inversely proportional to its exponential
moving average (EMA) response time. On every query the group picks a resolver
at random using those weights, so faster resolvers receive proportionally more
traffic. When a resolver fails (error, SERVFAIL with `servfail-error`, empty
response with `empty-error`, REFUSED, or NOTIMP), it is penalized by recording
`max(elapsed, failure-penalty)` as its effective RTT for that attempt, and the
query is retried with one of the remaining resolvers. The EMA naturally down-weights
stale history, so a resolver that recovers after a bad period regains its share
of traffic within tens of subsequent fast responses.

## Configuration

```toml
[groups.load-balanced]
type       = "load-balance"
resolvers  = ["upstream-a", "upstream-b", "upstream-c"]

# Optional. Seconds used as a floor penalty for failed resolvers (default 5).
failure-penalty = 5

# Optional. Treat SERVFAIL responses as failures and retry (default false).
servfail-error = false

# Optional. Treat empty responses as failures and retry (default false).
empty-error = false
```

See `cmd/routedns/example-config/load-balance.toml` for a working example.

## Changes

- `loadbalance.go` — new `LoadBalance` resolver group
- `loadbalance_test.go` — unit tests covering weighted selection, retry on
  error, penalty options, SERVFAIL/empty-error options, and all-fail behaviour
- `cmd/routedns/main.go` — wire up `load-balance` case in `instantiateGroup`
- `cmd/routedns/config.go` — add `failure-penalty` field to `group` struct;
  minor alignment fixes
- `cmd/routedns/example-config/load-balance.toml` — example configuration
- `doc/configuration.md` — document the new group type
- `README.md` — mention `load-balance` in the algorithm list
- `resolver.go` — extract shared `isSuccessResponse` package-level helper,
  consolidating identical logic that was duplicated across `failrotate`,
  `failback`, and `random`. This also fixes a latent bug in all three: a `nil`
  response (resolver returned nothing) was always treated as success regardless
  of `empty-error`. The original code had `return true // !r.opt.EmptyError` —
  the commented-out intent was never wired up. The correct behaviour,
  `return !emptyError`, is now applied consistently across all four group types.
- `failrotate.go`, `failback.go`, `random.go` — delegate to shared helper


cmd/routedns/config.go
```diff
@@ -120,9 +120,10 @@ type group struct {
 	EDNS0Data  []byte                  `toml:"edns0-data"`  // EDNS0 modifier option data
 
 	// Failover/Failback options
-	ResetAfter    *int `toml:"reset-after"`    // Time in seconds after which to reset resolvers in fail-back and random groups, or 0 to reset immediately, default 60.
-	ServfailError bool `toml:"servfail-error"` // If true, SERVFAIL responses are considered errors and cause failover etc.
-	EmptyError    bool `toml:"empty-error"`    // If true, empty responses are considered errors and cause failover etc.
+	ResetAfter     *int `toml:"reset-after"`     // Time in seconds after which to reset resolvers in fail-back and random groups, or 0 to reset immediately, default 60.
+	ServfailError  bool `toml:"servfail-error"`  // If true, SERVFAIL responses are considered errors and cause failover etc.
+	EmptyError     bool `toml:"empty-error"`     // If true, empty responses are considered errors and cause failover etc.
+	FailurePenalty *int `toml:"failure-penalty"` // Time in seconds used as a floor penalty for failed resolvers in load-balance groups, default 5.
```
folbricht: If you use `uint` here, non-pointer, that would simplify the code that reads it below. No nil-check, no negativity-check.

doc/configuration.md
```diff
@@ -623,6 +624,36 @@ resolvers = ["cloudflare-dot-1", "cloudflare-dot-2", "google-dot"]
 
 Example config files: [random-resolver.toml](../cmd/routedns/example-config/random-resolver.toml)
 
+### Load-Balance group
+
+This group picks one resolver for each query using weighted random selection. Resolvers with lower average response times receive more traffic. If a resolver fails, the request is retried with another resolver and the failed resolver receives a response-time penalty, reducing how often it is selected later.
+
+Unlike the [Fastest group](#fastest-group), this group does not send each query to every resolver.
+
+On startup, all resolvers have equal weight. If the selected resolver fails or times out, the group tries the same query with another resolver until one succeeds or all resolvers have been tried.
```
folbricht: It's not clear from this description how it differs from other failover groups like https://github.com/folbricht/routedns/blob/master/doc/configuration.md#fail-rotate-group

loadbalance.go
```diff
@@ -0,0 +1,182 @@
...
+
+type loadBalanceStats struct {
+	rttEMA float64 // exponential moving average in microseconds
+	count  int
+}
+
+// LoadBalanceOptions contain settings for the load-balancing resolver group.
+type LoadBalanceOptions struct {
+	// Duration used as the response-time penalty for failed lookups.
+	FailurePenalty time.Duration
```
folbricht, line 48: Is 0 a valid value for the penalty?

folbricht:
Thanks for that. I'm not quite done with the review, but should have time next week to dive deeper.

folbricht:
### Default behavior over-weights fast-failing resolvers

In `updateOnFailure`, when no penalty is applied — which is the **default** (`failure-penalty = 0`), and also every first failure before the consecutive-failure threshold — the failed call's *elapsed* time is blended directly into the EMA:

```go
penalize := r.opt.FailurePenalty > 0 && fails >= loadBalancePenaltyThreshold
d := elapsed
if penalize {
    d = r.opt.FailurePenalty
}
...
us := float64(d.Microseconds())   // EMA blended with this
```

Most real failures return *fast* — `REFUSED`/`NOTIMP`, connection-refused, SERVFAIL. A fast failure produces a small `us`, which **lowers** the EMA, which **raises** the resolver's weight (`weight = initialRTT / rttEMA`). So a dead resolver that rejects connections instantly ends up with the *highest* weight and is selected *most often*. With the 1µs floor, weight can reach `100000/1`; a ~0.5ms refused response yields weight ~200 vs ~5 for a healthy 20ms resolver — roughly 40× the traffic to the dead node.

Correctness is preserved (the query is retried and the client still gets an answer), but it inverts the stated goal. The docs say failing resolvers' "weight is suppressed via EMA," yet suppression only happens when failures are *slow* (timeouts). Out of the box, fast failures do the opposite, and every query keeps paying a wasted round trip + retry to the dead resolver. The optional penalty fixes it, but it's off by default and needs **two consecutive** failures — a flapping resolver whose successes reset `consecFails` never reaches the threshold.

Suggested fixes (any one):

- On a non-penalized failure, leave the EMA unchanged (don't record RTT at all), or
- Only let a failure move the EMA *upward*: `us = max(us, currentEMA)`, or
- Default `failure-penalty` to a sensible non-zero value and/or penalize from the first failure.

folbricht:
### Re: the `nil` response change — I'd suggest keeping the existing behavior

The refactor that extracts `isSuccessResponse` into a shared helper is a nice cleanup, but I don't think the `nil`-handling change is a bug fix — the original `return true` was deliberate, and flipping it changes behavior for `fail-rotate`/`fail-back`/`random` (not just the new group).

**History.** The `nil → true` short-circuit predates `empty-error` (it was `return a == nil || ...` originally). When `empty-error` was added, the `return true // !r.opt.EmptyError` line in #497 was reviewed: the comment documents the branch that was *intentionally not* taken. From [that review thread](https://github.com/folbricht/routedns/pull/497):

> I left the comment in to mean that this *would've* been an error if `empty-error` was set, but *isn't* because `nil` (as opposed to `[]`) is "deliberately empty".

and #500 then added `NilResponseWithEmptyError` asserting `require.True(...)` specifically to lock that in. This PR flips that assertion.

**Why it matters in practice.** `(nil, nil)` is a deliberate signal in RouteDNS, distinct from an empty answer section:

- `DropResolver` — "returns nil to signal to the listener to drop this request"
- `RateLimiter` — returns `(nil, nil)` to drop when the limit is exceeded and no `limit-resolver` is set

When either sits inside a `fail-rotate`/`fail-back`/`random`/`load-balance` group with `empty-error = true`, treating `nil` as a failure makes the group **fail over to another upstream and answer the query**, bypassing the intended drop/rate-limit. With the current `return true`, the drop is honored.

**Suggested change** — keep `nil` as success and revert the test:

```go
func isSuccessResponse(a *dns.Msg, servfailError, emptyError bool) bool {
	if a == nil {
		// A nil response is a deliberate "drop" signal (e.g. from the drop or
		// rate-limiter resolvers), distinct from an empty answer section. Treat
		// it as success so empty-error doesn't fail over and bypass the drop.
		return true
	}
	...
```

and in `failback_test.go`:

```go
t.Run("NilResponseWithEmptyError", func(t *testing.T) {
	fb := newFB(FailBackOptions{EmptyError: true})
	require.True(t, fb.isSuccessResponse(nil))
})
```

If `load-balance` wants different `nil` semantics it could handle that in its own wrapper, but the three existing groups shouldn't change behavior on upgrade.

folbricht:
### Question on recovery time for suppressed resolvers (non-blocking — checking intended behavior)

Not a blocker, just want to confirm this is the intended trade-off.

Because selection probability is proportional to weight, a heavily-suppressed resolver is picked rarely, which means its EMA also recovers slowly in query terms. With a 5s penalty its weight drops to roughly `100000 / 5000000 ≈ 0.02` versus ~5 for a healthy 20ms peer — about a 1-in-500 selection chance in a 3-resolver pool. Recovering the EMA back down to ~20ms at `alpha = 0.1` takes ~50 *successful* selections, but those are spread across thousands of queries since the resolver is rarely chosen.

So the doc's "regains its share of traffic within tens of subsequent fast responses" is accurate in terms of samples, but in wall-clock/query terms a resolver knocked down by a transient bad patch can stay suppressed much longer, and there's no exploration floor (e.g. epsilon-greedy / minimum weight) to speed reconvergence.

Is the slow recovery the intended behavior? If so, it might be worth softening the doc wording so operators don't expect fast recovery. (This interacts with the penalty design in my earlier comment.)

folbricht:
### Question on weight concentration vs. load spreading (non-blocking — checking intended behavior)

Also non-blocking, just confirming the design intent.

`weight = initialRTT / rttEMA` is linear and uncapped, and with the 1µs RTT floor a very fast resolver can reach weight ~`100000`. So if the pool mixes a microsecond-latency resolver (e.g. a local cache or static resolver) with 20ms network resolvers (weight ~5), the fast one captures ~99.99% of traffic rather than load being spread.

That may be exactly what's wanted given the group is documented as "preferring resolvers with lower average response times" — but it's effectively "prefer-fastest" rather than "balance load," which the name `load-balance` might lead operators to expect. If even distribution among comparably-healthy resolvers is desired, a weight cap or a gentler function (e.g. `sqrt`) would behave closer to the name.

Is the strong concentration on the fastest resolver intended? If so, a one-line note in the docs would set expectations.

loadbalance.go
```diff
@@ -0,0 +1,211 @@
...
+			return a, nil
+		}
+
+		log.With("resolver", resolver.String()).Debug("resolver returned failure",
+			"error", err)
+		r.metrics.failure.Add(resolver.String(), 1)
+		r.metrics.failover.Add(1)
+		penalized := r.updateOnFailure(idx, elapsed)
+		if penalized {
+			Log.Info("penalizing resolver", slog.Group("details",
```
folbricht: This logs at `Info` on every penalized query — the guard is `penalized` (`fails >= threshold`), not a state transition — so a sustained outage on one upstream would emit an Info line per failed attempt. `fail-rotate`/`fail-back` only log at `Info` on the actual failover transition. Suggest dropping this to `Debug` to avoid log spam under a persistent failure:

```suggestion
			Log.Debug("penalizing resolver", slog.Group("details",
```

folbricht:
@emlimap If you point you AI tool at the above comments it should take care of everything.

emlimap:
I have got AI to address those comments. Also got it do a review as well where it found and adressed an issue. Below is the summary of changes it did.

---

## Blocking issues

### Fast-failing resolvers boosted under default settings

With `failure-penalty = 0`, a resolver that fails instantly (~1 µs) got a
near-zero EMA, giving it enormous weight — the opposite of the intended
suppression.

**Fix (`loadbalance.go:updateOnFailure`):**
- **First observation** seeded at `max(elapsed, defaultLoadBalanceInitialRTT)`,
  flooring to 100 ms so a fast-failing resolver starts at the same neutral
  weight (1.0) as unprobed ones.
- **Subsequent failures** use an upward-only EMA clamp: the EMA can only increase
  on failure, preventing fast failures from lowering the EMA and boosting weight.

```go
// count == 0
s.rttEMA = max(us, float64(defaultLoadBalanceInitialRTT.Microseconds()))

// count > 0
newEMA := alpha*us + (1-alpha)*s.rttEMA
if penalize || newEMA > s.rttEMA {
    s.rttEMA = newEMA
}
```

---

### nil response handling — revert EmptyError interaction

`(nil, nil)` is a deliberate signal in RouteDNS (e.g. `DropResolver`) and must
always be treated as success. A prior commit incorrectly retried it when
`EmptyError = true`.

**Fix:** `isSuccessResponse` now returns `true` unconditionally for nil,
regardless of `EmptyError`. `EmptyError` only applies to non-nil responses with
no answer records.

Tests updated:
- `failback_test.go`: `NilResponseWithEmptyError` → expects success (was failure)
- `loadbalance_test.go`: `"Enabled"` split into `Enabled_NilIsSuccess` and
  `Enabled_EmptyMsgIsRetried`

---

## Non-blocking / acknowledged

**Penalty log level:** Changed `Log.Info` → `Log.Debug` at `loadbalance.go:112`
to avoid per-query log spam during sustained failures.

**Recovery time for suppressed resolvers:** Known trade-off of the EMA approach.
The upward-only clamp means recovery requires genuine successes, not just
failures at normal latency. Consistent with the docs — no change.

**Weight concentration:** Fast resolvers attracting more traffic is intentional.
No change.

---

## Additional fixes (from automated review, not raised by maintainer)

**Query mutation on retry (`loadbalance.go:102`):** The original code passed the
same `*dns.Msg` pointer to each resolver in the retry loop. Modifiers like
`EDNS0Modifier` mutate in-place, so subsequent retries received an already-
mutated query. Fixed by calling `q.Copy()` before each resolver invocation.

folbricht:
Thank you

PR #562, opened 2026-06-01
Fast-retry netns listener start to handle non-atomic ip netns add

folbricht:
## Problem

Reported in #522 (review comment): after a reboot, a "namespace creator" service that creates a netns and immediately issues a DNS query races RouteDNS's listener bind and fails every time, unless an arbitrary sleep is inserted.

Root cause: `ip netns add NAME` is **not atomic**. It first creates a mode-`000` placeholder file at `/run/netns/NAME` (which is what fires the inotify `CREATE` event the supervisor reacts to), then a moment later `unshare(CLONE_NEWNET)` + bind-mounts the real namespace onto that path. In the gap between the two:

- opening the path fails with `EACCES` for a non-root process (the placeholder is mode `000`; `CAP_SYS_ADMIN` does not bypass DAC), or `EINVAL` once `setns` sees a non-namespace fd.

Crucially, **no inotify event marks the bind mount completing** — `CREATE`, `OPEN`, and `CLOSE_NOWRITE` all fire on the placeholder *before* the mount, which is why watching `CLOSE` instead of `CREATE` didn't help. There is no "ready" signal to wait for.

The supervisor reacted to `CREATE`, tried to bind, hit `EACCES`, and then retried on a flat 1s interval — leaving the listener down for up to a second. A creator service that gives up and deletes the namespace within that window loses the race.

## Fix

Retry the listener start with an exponential backoff from a short initial interval (`25ms`) up to the existing 1s cap, instead of a flat 1s. The backoff resets on every namespace present/absent transition so each fresh appearance retries fast again. The namespace becomes usable within a millisecond or two of `CREATE`, so the listener now binds within tens of milliseconds rather than up to a second — comfortably before a creator service issues its first query.

## Changes

- `cmd/routedns/main.go`: add `netnsInitialRetryInterval`, replace the flat retry with a `scheduleRetry` backoff helper used by both the build-failure and start-failure paths; reset backoff on present/absent transitions.
- `cmd/routedns/netns_supervisor_test.go`: add `TestNetNSSupervisorRetriesStartFailureFast` covering the placeholder-window `EACCES` race; update the existing build-retry test to set both intervals.

### DIFF
--- a/cmd/routedns/main.go
+++ b/cmd/routedns/main.go
@@ -1121,11 +1121,26 @@ func networkForIPVersion(base string, ipVersion int) string {
 	return base + strconv.Itoa(ipVersion)
 }
 
-// netnsRetryInterval is how long the supervisor waits before retrying a
-// listener that failed to build or start while its namespace is present. It
-// matches the retry cadence used for listeners that aren't bound to a
-// namespace. It is a var so tests can shorten it.
+// netnsRetryInterval is the upper bound on how long the supervisor waits
+// before retrying a listener that failed to build or start while its namespace
+// is present. It matches the retry cadence used for listeners that aren't
+// bound to a namespace. It is a var so tests can shorten it.
 var netnsRetryInterval = time.Second
+
+// netnsInitialRetryInterval is the first (shortest) retry delay after a start
+// or build failure while the namespace is present. Retries back off
+// exponentially from here up to netnsRetryInterval.
+//
+// "ip netns add" is not atomic: it first creates a mode-000 placeholder file
+// (which is what fires the inotify CREATE event we react to), then a moment
+// later bind-mounts the real namespace onto it. In the gap between the two,
+// opening the path fails (EACCES on the placeholder for a non-root process, or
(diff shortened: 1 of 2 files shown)


Anuskuss:
Tbh I don't like it. Like I said, adding arbitrary delays sometimes wins you the race, but not always. I used to use 100 ms, then it stopped working. Then I used 10 ms, then it stopped working. Now I use 50 ms but have yet to reboot so who knows if it'll keep working. But if you can't find a better way to do it (e.g. check `/proc/self/mountinfo`?) I'm fine with it.
Anyway, it works at least. I'd suggest you don't issue a `WARN` in the first second though. The first second should be the "grace period", where it tries until it succeeds. After that, issue an `ERROR`/`WARN` because something is most likely wrong.

folbricht:
How about https://github.com/folbricht/routedns/pull/564? This avoids the timer-based retry loop and is more event-driven.

folbricht:
closing as https://github.com/folbricht/routedns/pull/564 supersedes it

PR #564, opened 2026-06-01
Detect netns readiness via mount events instead of timed retries

folbricht:
## Problem

Reported in #522 and addressed with timed retries in #562: a "namespace creator" service that runs `ip netns add` and immediately issues a DNS query races RouteDNS's listener bind. Any retry-interval approach wins or loses that race depending on timing (see discussion in #562).

There are actually two distinct windows where the listener isn't bound yet:

1. **Placeholder gap**: `ip netns add` is not atomic. It creates a mode-000 placeholder file (which fires the inotify CREATE event the watcher reacts to), then bind-mounts the real namespace onto it. Binds attempted in the gap fail with `EACCES`/`EINVAL`.
2. **Missing directory**: after a reboot, `/var/run/netns` itself doesn't exist until the first `ip netns add` creates it. The watcher polled for the directory every 5 seconds, so the first namespace created after startup could go unnoticed for up to 5s — a much larger window than the placeholder gap.

## Fix

Replace both timing-based waits with event-driven ones, eliminating retry intervals entirely:

- **`WaitNetNSReady`** (new in `netns_watch_linux.go`): after the namespace file appears, the supervisor waits for it to become an actual nsfs mount, checked with `statfs` + `NSFS_MAGIC`. The wait sleeps on `poll(POLLPRI)` on `/proc/self/mountinfo`, which the kernel signals whenever the mount table changes — so the wakeup happens at the moment the bind mount lands. This is the `/proc/self/mountinfo` approach suggested in the #562 discussion.
- **`waitForDir`**: when `/var/run/netns` doesn't exist, watch its parent directory with inotify for its creation instead of polling every 5s.

Both have fallbacks (timeout, slow polling) for pathological cases, but in normal operation no timer ever fires and no warning is logged while a namespace is being set up.

### Bounded re-check (added after testing feedback)

The mountinfo poll wakeup is not guaranteed in every environment: when RouteDNS runs inside its own mount namespace (the bundled systemd unit uses `DynamicUser=true`, which sandboxes the service), the namespace mount made by `ip` on the host has to *propagate* into RouteDNS's mount namespace, and on some kernels that propagation doesn't wake mountinfo pollers. Deleting the namespace file (a creator service cleaning up after a failed query) changes no mount state at all, so it can't wake the poll either.

Each poll is therefore capped at **10ms**, after which the path is re-checked. Where the wakeup works the listener still binds in ~1ms; where it doesn't, the worst case is ~10ms — bounded, and not a tunable race. A namespace deleted mid-wait is detected within the same bound instead of stalling until the 5s timeout.

## Result

Verified end-to-end with a real `ip netns add` + immediate query (the reported scenario), including the post-reboot case where `/run/netns` doesn't exist when RouteDNS starts:

```
22:06:05.219  netns present, starting listener   <- ip netns add
22:06:05.220  starting listener  proto=udp addr=127.0.0.1:5399
22:06:05.421  received query / answered          <- creator's first query
```

The listener binds **~1ms** after `ip netns add`, before the creator service can issue its first query, with no warnings logged. Rapid delete/recreate cycles behave the same.

## Changes

- `netns_watch_linux.go`: add `WaitNetNSReady` (statfs/NSFS_MAGIC + mountinfo poll); replace the 5s directory poll in `ensureWatch` with a parent-directory inotify watch
- `netns_watch_other.go`: non-Linux stub
- `cmd/routedns/main.go`: supervisor waits for namespace readiness before starting the listener
- Tests for readiness detection (immediate, missing, timeout, mount-wakeup [root-only]), directory-creation detection, and supervisor gating

Supersedes #562.

### DIFF
--- a/cmd/routedns/main.go
+++ b/cmd/routedns/main.go
@@ -1127,6 +1127,16 @@ func networkForIPVersion(base string, ipVersion int) string {
 // namespace. It is a var so tests can shorten it.
 var netnsRetryInterval = time.Second
 
+// netnsReadyTimeout is the longest the supervisor waits for a namespace that
+// just appeared to become fully set up (see waitNetNSReady) before trying to
+// start the listener anyway. Normally the wait completes within a few
+// milliseconds; the timeout only matters in pathological cases such as a
+// stray non-namespace file under /var/run/netns.
+var netnsReadyTimeout = 5 * time.Second
+
+// waitNetNSReady is rdns.WaitNetNSReady, indirected so tests can stub it out.
+var waitNetNSReady = rdns.WaitNetNSReady
+
 // superviseNetNSListener gates a listener on the presence of a network
 // namespace. When the namespace appears, build is called and the resulting
 // listener is started; when it disappears, the listener is stopped. A fresh
@@ -1184,6 +1194,15 @@ func runNetNSSupervisor(id, nsName string, events <-chan rdns.NetNSState, build 
 					continue
 				}
 				rdns.Log.Info("netns present, starting listener", "netns", nsName, "id", id)
+				// "ip netns add" creates the namespace file before mounting
+				// the actual namespace onto it, and the inotify event fires on
+				// the former. Wait for the mount so the listener doesn't try
+				// to bind into the placeholder. On error (namespace deleted
+				// again, timeout, ...) still attempt the start: a failure
+				// there is visible to the user and retried on a timer.
+				if err := waitNetNSReady(nsName, netnsReadyTimeout); err != nil {
+					rdns.Log.Warn("netns not ready", "netns", nsName, "id", id, "error", err)
+				}
 				launch()
 			case rdns.NetNSAbsent:
 				present = false
--- a/cmd/routedns/main.go
+++ b/cmd/routedns/main.go
@@ -1132,7 +1132,7 @@ var netnsRetryInterval = time.Second
 // start the listener anyway. Normally the wait completes within a few
 // milliseconds; the timeout only matters in pathological cases such as a
 // stray non-namespace file under /var/run/netns.
-var netnsReadyTimeout = 5 * time.Second
+var netnsReadyTimeout = time.Second
 
 // waitNetNSReady is rdns.WaitNetNSReady, indirected so tests can stub it out.
 var waitNetNSReady = rdns.WaitNetNSReady
--- a/cmd/routedns/main.go
+++ b/cmd/routedns/main.go
@@ -1124,15 +1124,15 @@ func networkForIPVersion(base string, ipVersion int) string {
 // netnsRetryInterval is how long the supervisor waits before retrying a
 // listener that failed to build or start while its namespace is present. It
 // matches the retry cadence used for listeners that aren't bound to a
-// namespace. It is a var so tests can shorten it.
-var netnsRetryInterval = time.Second
+// namespace.
+const netnsRetryInterval = time.Second
 
 // netnsReadyTimeout is the longest the supervisor waits for a namespace that
 // just appeared to become fully set up (see waitNetNSReady) before trying to
 // start the listener anyway. Normally the wait completes within a few
 // milliseconds; the timeout only matters in pathological cases such as a
 // stray non-namespace file under /var/run/netns.
-var netnsReadyTimeout = time.Second
+const netnsReadyTimeout = time.Second
 
 // waitNetNSReady is rdns.WaitNetNSReady, indirected so tests can stub it out.
 var waitNetNSReady = rdns.WaitNetNSReady
--- a/cmd/routedns/netns_supervisor_test.go
+++ b/cmd/routedns/netns_supervisor_test.go
@@ -68,9 +68,7 @@ func TestStopNetNSListenerRetriesUntilStopped(t *testing.T) {
 
 		stopNetNSListener("test", f, startErr)
 
-		if c := f.calls(); c < 2 {
-			t.Fatalf("expected Stop() to be retried, got %d call(s)", c)
-		}
+		require.GreaterOrEqual(t, f.calls(), 2, "expected Stop() to be retried")
 	})
 }
 
@@ -84,9 +82,7 @@ func TestStopNetNSListenerStopsImmediately(t *testing.T) {
 
 		stopNetNSListener("test", f, startErr)
 
-		if c := f.calls(); c != 1 {
-			t.Fatalf("expected exactly 1 Stop() call, got %d", c)
-		}
+		require.Equal(t, 1, f.calls(), "expected exactly 1 Stop() call")
 	})
 }
 
--- a/cmd/routedns/netns_supervisor_test.go
+++ b/cmd/routedns/netns_supervisor_test.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"sync"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	rdns "github.com/folbricht/routedns"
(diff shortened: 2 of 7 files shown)


Anuskuss:
Assuming the namespace is called `test`, this first `inotifywait`s `/run` until `/run/netns` appears (didn't even know you had to do that), then `inotifywait`s `/run/netns` until `test` appears, and then `poll`s `/proc/self/mountinfo` until `test` (or any `nsfs`?) appears? That got complicated real fast.
Either way, if that's all true and timers are only used as fallbacks, then I have no complaints.

For whatever reason the 5 second timer started (which is of course too long because my `ns-creator` has already cleaned up the namespace at that point):
```
Jun 01 22:58:12 routedns[180171]: time=2026-06-01T22:58:12.431+02:00 level=INFO msg="starting listener" id=udp protocol=udp4 addr=:5300
Jun 01 22:58:12 routedns[180171]: time=2026-06-01T22:58:12.431+02:00 level=INFO msg="starting listener" id=tcp protocol=tcp4 addr=:5300
Jun 01 22:58:12 routedns[180171]: time=2026-06-01T22:58:12.431+02:00 level=WARN msg="netns not present, waiting" netns=test id=udp_test
$ systemctl start ns-creator.service
Jun 01 22:58:18 routedns[180171]: time=2026-06-01T22:58:18.769+02:00 level=INFO msg="netns present, starting listener" netns=test id=udp_test
Jun 01 22:58:18 routedns[180171]: time=2026-06-01T22:58:18.769+02:00 level=INFO msg="starting listener" id=udp_test protocol=udp4 addr=:53
$ systemctl restart ns-creator.service
Jun 01 22:58:35 routedns[180171]: time=2026-06-01T22:58:35.609+02:00 level=WARN msg="netns gone, stopping listener" netns=test id=udp_test
Jun 01 22:58:35 routedns[180171]: time=2026-06-01T22:58:35.670+02:00 level=INFO msg="netns present, starting listener" netns=test id=udp_test
Jun 01 22:58:40 routedns[180171]: time=2026-06-01T22:58:40.673+02:00 level=WARN msg="netns not ready" netns=test id=udp_test error="netns \"test\" does not exist"
Jun 01 22:58:40 routedns[180171]: time=2026-06-01T22:58:40.673+02:00 level=WARN msg="netns gone, stopping listener" netns=test id=udp_test
Jun 01 22:58:40 routedns[180171]: time=2026-06-01T22:58:40.673+02:00 level=WARN msg="error stopping listener" id=udp_test error="dns: server not started"
Jun 01 22:58:40 routedns[180171]: time=2026-06-01T22:58:40.673+02:00 level=INFO msg="starting listener" id=udp_test protocol=udp4 addr=:53
```

folbricht:
Your log exposed a gap: the mountinfo poll wakeup isn't reliable when RouteDNS is sandboxed in its own mount namespace (`DynamicUser=true` in the unit file), and a deleted namespace file never generates a mount event at all. Pushed a fix: the poll is now capped at 10ms and the namespace path re-checked after each, so the wakeup is an optimization rather than a requirement. Worst case the listener binds ~10ms after ip netns add; the deleted-namespace case you hit now resolves in ~10ms instead of 5s. Would you mind giving the updated branch a try?

Anuskuss:
Hm it works but I'm still kinda unsatisfied. What about dropping all the extra code and just watching `/proc/self/mountinfo`? I dislike that it first has to watch `/run` for `netns`, which a) might not ever happen and b) will fire very often (e.g. it'll create 30 events, that RouteDNS has to handle, when creating a new namespace). I think that  `/proc/self/mountinfo` will get updated less.
Or instead of trying to handle all edge cases what about just `Error`ing when `/run/netns` doesn't exist? That way I could make sure to create it myself because my setup is kinda special (e.g. `ExecStartPre=+-mkdir /run/netns`). Then afterwards RouteDNS would simply watch `/run/netns` and wait for the namespace. Then it would try it bind it once every 10 (?) ms until it succeeds, up to 1 second and without warnings. After 1 second it could either retry every 1 second until a timeout is reached (e.g. 5) or just `Error` (I prefer this because 1 second is plenty of time and it's unlikely that it will take >1s but <5s). Thoughts?

FWIW for me the time between namespace creation and DNS access is (at the earliest) ~407 ms. However, if I shuffle things around (this doesn't change the behaviour) it's ~551 ms. So even 100 ms steps would work for me. And I'm all for keeping the code simple so I'm in favor of dropping this PR and going back to #562.

folbricht:
Watching only `/proc/self/mountinfo` doesn't work, unfortunately, your earlier log shows why. When RouteDNS is sandboxed in its own mount namespace (DynamicUser=true in the unit), mounts propagating in from the host don't reliably wake mountinfo pollers, and deleting a namespace file changes no mount state at all, so a mountinfo-only watcher would just reintroduce the 5s stall you hit. It also fires on every mount change system-wide and would need parsing to map mounts back to namespace names (more code, not less). The `/run` watch you noticed is temporary: it only exists until /run/netns appears once, then it's closed.

Erroring when `/run/netns` doesn't exist would break the normal post-reboot case (the directory doesn't exist until the first ip netns add), and requiring a mkdir in ExecStartPre pushes a boot-ordering requirement onto everyone , the opposite of what this feature is for.

That said, you're right that there was too much machinery. The bounded re-check had already made the mountinfo poll nearly pointless (it saved at most ~9ms), so I dropped it entirely: after the inotify event, readiness is now just a statfs check every 10ms until the namespace mount lands (7d107ab). Same behavior, same worst case, ~25 fewer lines and no procfs polling tricks. That's effectively your "try every 10ms" suggestion, just gated on the inotify event and without giving up after a fixed timeout, since the namespace can legitimately appear at any time.

If you try this one, please include logs if you see anything weird. I'm not currently able to spend much (any) time on really testing this.

Anuskuss:
> Erroring when `/run/netns` doesn't exist would break the normal post-reboot case

Yeah but it's not RouteDNS's job to make sure it exists. I think watching `/run` is way too heavy (it will literally fire hundreds of times during normal use) and beyond the scope of what a DNS server should have access to. Most people using network namespaces in RouteDNS will naturally order it *after* the namespace exists. I guess I'm the exception and that's why I suggested the `ExecStartPre` (it was more of a suggestions for people finding this issue rather than you adding it to the official `.service`). Again, please remove that code - it doesn't belong here and is a waste of resources (like I said in an earlier comment, sometimes I don't even create the namespace so RouteDNS will just eat CPU cycles).

> readiness is now just a statfs

Why? Is there a difference between the error from binding the "000 placeholder" and the user not having the necessary permission (e.g. `CAP_SYS_ADMIN`)? Why don't you just try to bind and handle the error?

folbricht:
**On "why not just bind and handle the error":** your question is the answer. There is no reliable difference between those errors. Without root, opening the mode-000 placeholder fails with `EACCES`, identical to a genuine permission problem. With root, the open succeeds and setns fails with `EINVAL`, identical to any stray non-namespace file. So an error-driven retry can't tell transient from permanent: it either retries every error forever (and spams the log on real misconfigurations) or gives up on errors that were about to resolve. statfs is what makes it  tractable: checking for NSFS_MAGIC answers "is the real namespace mounted yet" in one cheap syscall, needs no permissions on the file, and skips the LockOSThread/setns/socket churn a real bind attempt costs. Bind errors are still handled after that; the statfs wait just keeps an expected transient state from producing noise.

**On the /run watch:** small correction first. It only exists while `/run/netns` is missing; the moment the directory appears, that fd is closed and it never fires again. In normal use it fires zero times. In your never-create-the-namespace case, each event costs one `read()` wakeup and one `stat()`, so the CPU cost is effectively zero even at hundreds of events. That said, I take the scope point about holding a watch on a busy system directory indefinitely.

**Middle ground:** I can replace the parent watch with a plain stat of  `/run/netns` once a second while it's missing. No watch on `/run`, one syscall per second only while the directory doesn't exist, and the post-boot window shrinks from the old 5s to ~1s instead of disappearing entirely. The trade-off is that a namespace created within ~1s of the directory itself appearing could be noticed up to a second late, and with your creator querying at ~400ms you'd be the user most likely to hit that (only on the very first namespace after boot, with your `ExecStartPre` mkdir it never happens at all). Erroring out when the directory is missing is something I'd rather avoid since it reintroduces a boot-ordering requirement for everyone else. Would the 1s poll work for you?

Anuskuss:
> Erroring out when the directory is missing is something I'd rather avoid since it reintroduces a boot-ordering requirement for everyone else.

Like I said, the majority of people running namespaces in RouteDNS will want to order RouteDNS *after* the namespace creation anyway. One could argue that watching `/run/nets` goes too far as well but watching `/run` definitely does. Even if the cost is negligible, what if I need the namespace only in let's say May? Then it watches `/run` (probably hundreds of thousand events at that point) or polls (literally millions of times) for no real benefit. A clean error after reboot ("/run/netns doesn't exist, will not watch for namespace creation") (that can be worked around by experiences users who decide to run RouteDNS *before* the namespace creation), an `inotifywait` on `/run/netns` (which admittedly could also fire a bunch of time, although that's unrealistic) and a 10 ms warning-less 1 second fast-retry will suffice. Everything beyond that is overkill and I don't want/need it.

folbricht:
That's fair enough. It's something the user can easily handle themselves. The `/run` watch is gone entirely now. The inotify watch on `/run/netns` is now set up once at startup, and if the directory doesn't exist the listener is disabled with a clear error (`"netns directory /var/run/netns does not exist; create it before starting routedns, or start routedns after the first ip netns add"`). Your `ExecStartPre=+mkdir -p /run/netns` trick is now the documented workaround for starting RouteDNS before any namespace exists.

So the current behavior is: one inotify watch on `/run/netns`, a 10ms warning-free statfs re-check after the create event until the namespace mount lands (up to 5s), and a clean error if the directory is missing at startup. No watch on `/run`, no polling, no timers in the normal path. I think that matches what you described. Mind giving the branch one more spin?

The 5s seem a little long, should I drop that to 1s?

Anuskuss:
> So the current behavior is: one inotify watch on `/run/netns`, a 10ms warning-free statfs re-check after the create event until the namespace mount lands (up to 5s), and a clean error if the directory is missing at startup. No watch on `/run`, no polling, no timers in the normal path. I think that matches what you described. Mind giving the branch one more spin?

Thanks, that's more reasonable now.

> The 5s seem a little long, should I drop that to 1s?

I think 1 second is plenty.

Anuskuss:
Tested and works.

folbricht:
Thank you. Should be good to go then.

PR #580, opened 2026-06-12
Use expiration-cache sharding for prefetch tracking

emlimap:
Saw the library got updated, so got Claude to investigate if it is useful and it did see some improvements so raising it as a PR. Please feel free to close if you think otherwise

## Summary

This updates the dedicated `prefetch` resolver to use the sharding support added
in `github.com/0xERR0R/expiration-cache` v0.2.0.

Both internal prefetch tracking caches now set `Options.Shards` automatically
based on `runtime.GOMAXPROCS(0)`. Single-core processes and cache capacities of
one remain unsharded.

## Rationale

`expiration-cache` uses an LRU internally, and `Get` updates recency. That means
prefetch cache reads still take an LRU lock. Under concurrent DNS traffic, the
dedicated `prefetch` group can hit these caches frequently:

- `nameCache.Get` / `nameCache.Put` for successful prefetch-eligible responses
- `queryCache.Put` once a query reaches the configured threshold
- `queryCache.Get` from the expiration callback when refreshing hot entries

v0.2.0 can split the cache across independent LRU shards, reducing lock
contention when different query names are handled concurrently on multi-core
systems.

## Benchmarks

`BenchmarkPrefetch*` drives `Prefetch.Resolve` concurrently (`-cpu=8`) and
compares a single-shard cache (previous behavior) against a GOMAXPROCS-sharded
cache. The benchmark uses an instant in-memory resolver to isolate the prefetch
group's own cache bookkeeping. Median of 5 runs on a 4-core host:

| Scenario | Unsharded | Sharded | In-bookkeeping delta |
|---|---|---|---|
| Hot keys (8 names, high contention) | 4033 ns/op | 3537 ns/op | ~12% lower |
| Spread (4096 names, low contention) | 4076 ns/op | 3045 ns/op | ~25% lower |

Allocations are unchanged (26 allocs/op) — sharding only splits the lock and
adds no per-query overhead.

**Scope of the win:** this is a sub-microsecond reduction in the prefetch
group's in-process bookkeeping, not an end-to-end speedup. Real queries are
dominated by upstream network/TLS latency (milliseconds), so this change is
not observable in user-facing query latency. It only relieves lock contention,
and only matters at high concurrent QPS on multi-core systems. Treat it as a
defensive scalability improvement that comes for free with the v0.2.0 upgrade,
not a latency optimization.

## Tradeoff

With sharding enabled, LRU eviction is per-shard rather than globally exact. For
prefetch tracking this is acceptable because the data is approximate popularity
bookkeeping, not authoritative DNS response storage.

## Changes

- Bump `github.com/0xERR0R/expiration-cache` to v0.2.0 (adds `Options.Shards`).
- Add `expirationCacheShards` helper for selecting a shard count.
- Configure sharding for both prefetch expiration caches.
- Refactor `NewPrefetch` to delegate to an internal `newPrefetchWithShards`
  constructor so the shard count can be forced in benchmarks; `NewPrefetch`
  keeps the existing auto-sizing behavior.
- Add a focused unit test for shard selection and concurrent `Resolve`
  benchmarks comparing sharded vs. unsharded.

## Testing

```sh
go test -run 'TestPrefetch|TestExpirationCacheShards' .
go test -run '^$' -bench 'BenchmarkPrefetch' -benchmem -cpu=8 .
```

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -454,6 +454,8 @@ Example config files: [cache.toml](../cmd/routedns/example-config/cache.toml), [
 
 While [Cache](#cache) has built-in prefetch capabilities, the dedicated `prefetch` group may be more appropriate for some use cases. It tracks the number of queries made within a time window and actively prefetches frequently requested records. While it actively sends queries in order to refresh a cache, it does not cache responses itself and relies on a cache upstream from it.
 
+On multi-core systems the internal tracking caches are automatically sharded (based on `GOMAXPROCS`) to reduce lock contention under concurrent query load. This is an internal optimization only; it does not change behavior or measurably affect end-to-end query latency.
+
 #### Configuration
 
 Prefetch groups are instantiated with `type = "prefetch"` in the groups section of the configuration.
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/folbricht/routedns
 go 1.25.0
 
 require (
-	github.com/0xERR0R/expiration-cache v0.1.0
+	github.com/0xERR0R/expiration-cache v0.2.0
 	github.com/BurntSushi/toml v1.6.0
 	github.com/RackSec/srslog v0.0.0-20180709174129-a4725f04ec91
 	github.com/cisco/go-hpke v0.0.0-20230407100446-246075f83609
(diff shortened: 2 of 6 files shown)


folbricht:
Thank you

PR #581, opened 2026-06-15
Improve load-balance group recovery, fairness, and observability

emlimap:
## Motivation

The `load-balance` group (#538) steers traffic by weighted random selection using an
exponential moving average (EMA) of each resolver's response time, with weight =
`100ms / EMA`. This works well in steady state, but the dynamics have three gaps:

1. **Slow recovery.** A resolver penalized by `failure-penalty` (or whose EMA was
   inflated by timeouts) only improves its average through successful samples. With a
   5s EMA its weight is ~0.02 vs the neutral 1.0, so it receives ~1% of traffic — and
   with alpha = 0.1 it needs 20+ successes to climb back. On a low-traffic server that
   can take hours after the upstream is already healthy again.
2. **Starvation.** The EMA floor is 1µs, so a very fast upstream (e.g. a LAN resolver
   at 1ms) gets weight 100 and effectively crowds out the others. Their stats go stale,
   and during cold start the group converges prematurely on whichever resolver
   happened to answer first.
3. **No visibility.** The EMA drives all routing decisions but isn't exported anywhere,
   making it hard to understand traffic skew or tune `failure-penalty`.

## Changes

All changes are internal to the group — **no new configuration options**, existing
configs behave the same in steady state.

- **Exploration floor (ε-greedy):** ~5% of picks select a resolver uniformly at random
  instead of by weight. This guarantees every resolver a minimum traffic share, keeps
  response-time stats fresh, lets penalized resolvers be re-measured (and recover), and
  avoids cold-start lock-in.
- **Fast recovery after failure streaks:** on the first success following a streak of
  2+ consecutive failures, the EMA is re-seeded to the freshly measured RTT instead of
  alpha-blended. A recovered upstream returns to full weight on its first successful
  probe rather than over thousands of queries.
- **Bounded weighting:** the RTT used for weight computation is floored at 1ms (max
  weight 100 vs neutral 1.0). The stored EMA is unchanged — the clamp applies only at
  selection time — so metrics and recovery still see true values.
- **Per-resolver RTT metric:** the current EMA (in microseconds) is exported via expvar
  as `routedns.router.<id>.rtt`, keyed by resolver, alongside the existing `route`,
  `failure`, `available`, and `failover` metrics.
- **`failover` metric accuracy:** the counter now only increments when another resolver
  is actually retried; the final resolver failing is the end of the request, not a
  failover. `failure` still counts every failure.

## Benefits

- A dead upstream is demoted within one or two failures (first 2s timeout alone drops
  its weight 20×) and, once healthy, is restored to full weight by a single successful
  exploration probe — no operator intervention, no waiting out a long decay.
- Slower-but-healthy resolvers keep receiving a measurable share of traffic, so their
  measurements stay current and the group adapts when relative performance shifts
  (e.g. an upstream's cache warming up). On startup the same mechanism ensures every
  resolver gets measured instead of the group locking onto whichever answered first.
- Operators can watch `routedns.router.<id>.rtt` to see exactly why traffic is
  distributed the way it is and to choose a sensible `failure-penalty`.

## Implementation notes

- Randomness still flows through the single injectable `randFloat` field, so existing
  deterministic tests are unchanged (a constant-0 override explores and picks index 0,
  matching prior behavior).
- The failure-streak read-and-reset uses `atomic.Swap`, keeping `consecFails` out of
  the mutex as before.
- Fast recovery triggers on any 2+ failure streak, including with `failure-penalty`
  disabled — intentional, since a timeout-inflated EMA benefits from the same reset.

## Testing

- Four new tests: exploration share for a heavily penalized resolver, EMA re-seeding
  after a penalty streak, weight-clamp bounding, and failover-counter accuracy. All use
  distinct group ids to avoid expvar collisions.
- `go test -race ./...` passes (336 tests), `gofmt` and `go vet` clean.

## Documentation

The Load-Balance section of `doc/configuration.md` now describes the exploration share,
bounded weighting, fast recovery, and the new `rtt` metric.

### DIFF
--- a/loadbalance.go
+++ b/loadbalance.go
@@ -2,6 +2,7 @@ package rdns
 
 import (
 	"errors"
+	"expvar"
 	"math/rand"
 	"sync"
 	"sync/atomic"
@@ -19,6 +20,17 @@ const (
 	// Two transient failures in a row are unlikely to be coincidental; apply the
 	// penalty only then to avoid suppressing a resolver that had a single slow response.
 	loadBalancePenaltyThreshold = 2
+	// Fraction of picks made uniformly at random instead of by weight. Exploration
+	// keeps every resolver's stats fresh, guarantees a minimum traffic share so that
+	// penalized or slow resolvers can recover (and be re-measured) rather than being
+	// starved indefinitely, and avoids cold-start lock-in onto whichever resolver
+	// happened to be probed first.
+	defaultLoadBalanceExploration = 0.05
+	// Floor for the EMA used when computing weights. Without it a resolver that
+	// responds in microseconds would get a weight thousands of times the neutral
+	// weight, starving slower or unprobed resolvers. Clamping to 1ms caps the
+	// maximum weight at 100 (vs. the neutral 1.0).
+	defaultLoadBalanceMinimumWeightRTT = time.Millisecond
 )
 
 // LoadBalance is a resolver group that prefers resolvers with lower average
@@ -30,11 +42,14 @@ type LoadBalance struct {
 	stats     []loadBalanceStats
 	opt       LoadBalanceOptions
 	metrics   *FailRouterMetrics
+	// Per-resolver current rttEMA in microseconds, published under
+	// routedns.router.<id>.rtt keyed by resolver String().
+	rttVars   []*expvar.Float
 	randFloat func() float64
 }
 
 type loadBalanceStats struct {
-	rttEMA      float64      // exponential moving average in microseconds
+	rttEMA      float64 // exponential moving average in microseconds
 	count       int
 	consecFails atomic.Int32 // reset to 0 on success; never read under mu
 }
@@ -60,12 +75,19 @@ var _ Resolver = &LoadBalance{}
 
 // NewLoadBalance returns a new instance of a load-balancing resolver group.
 func NewLoadBalance(id string, opt LoadBalanceOptions, resolvers ...Resolver) *LoadBalance {
+	rtt := getVarMap("router", id, "rtt")
+	rttVars := make([]*expvar.Float, len(resolvers))
+	for i, resolver := range resolvers {
+		rttVars[i] = new(expvar.Float)
+		rtt.Set(resolver.String(), rttVars[i])
+	}
 	return &LoadBalance{
 		id:        id,
 		resolvers: resolvers,
 		stats:     make([]loadBalanceStats, len(resolvers)),
 		opt:       opt,
 		metrics:   NewFailRouterMetrics(id, len(resolvers)),
+		rttVars:   rttVars,
 		randFloat: rand.Float64,
 	}
 }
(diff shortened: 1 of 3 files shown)


loadbalance.go
```diff
@@ -172,16 +209,25 @@ func (r *LoadBalance) updateOnSuccess(idx int, d time.Duration) {
...
+	// Swap so the read-and-reset of the failure streak is a single atomic op.
+	wasPenalized := r.stats[idx].consecFails.Swap(0) >= loadBalancePenaltyThreshold
 	r.mu.Lock()
 	s := &r.stats[idx]
-	if s.count == 0 {
+	switch {
+	case s.count == 0:
 		s.rttEMA = us
-	} else {
+	case wasPenalized:
```
folbricht: The re-seed condition is broader than the comment claims. `wasPenalized` is derived from
`consecFails >= loadBalancePenaltyThreshold` alone, so this branch also fires when
`failure-penalty` is unset (the default), where `updateOnFailure` only ever moves the
EMA upward and nothing inflated it.

Concrete case with `LoadBalanceOptions{}`: a resolver with an honest 400ms EMA returns
two fast transient failures (connection refused in ~1ms). Those leave the EMA at 400ms
(the blended value 360ms is lower, so the upward-only update skips it), but consecFails
still reaches 2. The next success happens to be a 1ms cached answer, and this branch
hard-resets the EMA to 1ms instead of blending to ~360ms. With the new 1ms weighting
floor that resolver jumps from weight ~0.28 straight to the max of 100 and gets flooded
with traffic it actually serves at 400ms, until the EMA slowly climbs back.

Since `updateOnFailure` already knows whether it inflated the EMA (the `penalize` path,
or a timeout sample that raised it), could that fact be recorded per resolver and used
here instead of re-deriving it from the streak length? Then recovery stays immediate for
genuinely inflated EMAs and an established average is never wiped by a single sample.
The branch comment ("The EMA was inflated by the failure penalty") would then also be
accurate again.

loadbalance.go
```diff
@@ -109,7 +131,11 @@ func (r *LoadBalance) Resolve(q *dns.Msg, ci ClientInfo) (*dns.Msg, error) {
 		log.With("resolver", resolver.String()).Debug("resolver returned failure",
 			"error", err)
 		r.metrics.failure.Add(resolver.String(), 1)
-		r.metrics.failover.Add(1)
+		// Only count a failover when there is another resolver to fall back to;
+		// the last resolver failing is the end of the request, not a failover.
+		if len(remaining) > 1 {
```
folbricht: This makes `failover` diverge from the other groups rather than align with them.
FailRotate's `errorFrom` increments the counter on every failure, including the last or
only resolver (it wraps around), so after this change the same fully-failed request
reports N in a fail-rotate group but N-1 here.

The case I care about most: a load-balance group with a single resolver whose upstream
is dead now reports `failover=0` forever. Anyone alerting on
`routedns.router.<id>.failover` loses the one signal that flagged a completely broken
group. `failure` still counts, but existing dashboards keyed on `failover` will go
quiet without warning.

I would rather keep the unconditional increment here, and treat "failures that had a
fallback available" as a separate discussion if it is useful (either a new counter or a
documented semantics change applied consistently across the groups). If you have a
strong reason for the narrower definition, the doc change should at least call out that
load-balance counts differently from fail-rotate.

emlimap:
Addressed your PR comment. Also got Claude to do a review and addressed the issue it raised

folbricht:
Thank you

PR #603, opened 2026-08-07
Write the memory cache and blocklist files atomically

emlimap:
## Summary

The memory backend's cache-file persistence (`MemoryBackendOptions.Filename` /
`save-interval`) spends more CPU and does more syscalls than it needs to. This change
makes the save path ~4.5× faster and removes over 99% of its allocations, **without
changing the on-disk format** — files written before and after this change are
byte-identical.

This is aimed at the low-powered OpenWRT devices running RouteDNS, where the periodic `save-interval` write is a
noticeable CPU spike and the cache file usually lives on tmpfs.

No changes to the `CacheBackend` interface, the in-memory `lruCache` structure, the
query path, or the config surface.

## Changes

**1. Append-based JSON encoder for `serialize()`**

`serialize()` used `json.NewEncoder`, which reflects over every record and — because
`cacheAnswer` has a `MarshalJSON` method — re-parses that method's output through
`compact()` to strip whitespace. Every record was therefore encoded twice.

Records are now written by appending to a reused buffer. `Msg.PackBuffer` and
`base64.AppendEncode` reuse scratch space instead of allocating per record.

Note `PackBuffer` decides whether it can reuse the caller's buffer from its *length*,
not its capacity, and returns a length-truncated slice. Passing that slice straight back
in therefore reuses nothing — the wire buffer must be handed back as `wire[:cap(wire)]`,
keeping whichever buffer is larger. This is what takes the per-save allocations from
one-per-record down to a constant.

The output is **byte-for-byte identical** to what `encoding/json` produced; a test
asserts this directly against `json.NewEncoder` across plain, fully-populated, and
escape-requiring records. `MarshalJSON` now has no production callers but is kept
deliberately: it is the definition of the on-disk format, and the test uses it as the
oracle the append encoder is checked against. `deserialize` is unchanged, so old files
load and new files are readable by older builds.

Note the timestamps keep whatever zone offset `encoding/json` would have written rather
than being normalised to UTC — forcing UTC here would be a format change, and the
byte-identity test catches it.

**2. Buffered file I/O**

`serialize` wrote straight to `*os.File`, so encoding a cache issued at least one
`write` syscall per record — 1000+ syscalls for a 1000-record cache. Both the read
and write paths now go through a 64 KB `bufio` wrapper.

**3. Skip expired records on save and on load**

Expired entries were written to the file and loaded back into the LRU, where they
occupied capacity — potentially evicting live records via `resize()` — until the next
lookup or GC pass (up to `GCPeriod` later). They are now skipped in both directions.
This also shrinks the file and the load time.

**4. Narrow the lock held during save, and make the write atomic**

`writeToFile` held `b.mu` across file creation, encoding and all I/O, blocking every
query for that whole window. The lock now covers only the encode; file creation, the
final flush and the rename are outside it. Combined with (1), the window queries can
block on is substantially shorter.

The lock is not entirely free of I/O: `serialize` writes into a `bufio.Writer`, which
flushes to the file whenever it fills, so a large cache still performs some write
syscalls under `b.mu`. Buffering the whole encode in memory first would remove that,
but at a cost (tens of MB for a large cache) that is the wrong trade on the devices this
targets. The comment in `writeToFile` says so rather than claiming the lock is I/O-free.

Narrowing `b.mu` removes the mutual exclusion between concurrent `writeToFile` calls
that holding it for the whole function previously provided as a side effect — a
periodic `save-interval` write can overlap the one from `Close()`.

The save now writes to a uniquely-named temp file and renames it into place. That
rename is what makes overlapping saves safe: each writes its own file, and the result
is always one complete save rather than two interleaved. It also means an error or a
crash part-way through leaves the previous cache file intact instead of a truncated
one — previously `os.Create` truncated the real file before any new content was
written, so an interrupted save destroyed the cache.

`HTTPLoader.writeToDisk` already implemented this same temp-file-then-rename sequence,
so rather than adding a second copy it is extracted into `writeFileAtomic`
(`atomicfile.go`) and both callers now use it. The two had already drifted — the
blocklist version silently discarded `Flush` errors — and the "close before rename,
Windows needs it" constraint was recorded in only one of them.

Consolidating also fixes two latent bugs in the blocklist path: it discarded the errors
from both `Flush` and `Close`, so an ENOSPC during flush would rename a **truncated**
blocklist into place. The shared helper checks both, and additionally `fsync`s the temp
file before the rename so a machine crash can't leave the new name pointing at an empty
file. (The containing directory is not synced, so the rename itself can still be lost —
which leaves the previous file intact, the outcome the helper promises.)

A dedicated `saveMu` additionally serializes saves. This is not required for file
integrity (the rename already provides that); it avoids two saves redundantly encoding
the same cache at once, which on a weak core is exactly the CPU spike this PR is
trying to reduce. It is a separate lock from `b.mu`, which is held only for the encode
rather than for the whole save.

**5. A failed pack no longer aborts the whole save**

A record whose message fails to `Pack()` previously returned an error from
`serialize()` and abandoned the entire cache file. Such a record can't be restored
anyway, so it is now skipped and the rest of the cache is still written.

This is a deliberate behaviour change: a save can now succeed while silently omitting
records. To keep that visible without flooding the log on every `save-interval`, the
skips are reported as a single summary line with a count and the *first* error — for a
systematic failure the first one is the actionable one, where the last is arbitrary.

`serialize()` takes the logger to report through and returns the skip count rather than
deciding how to surface it, leaving `lruCache` free of a dependency on the package-global
`Log`. The memory backend increments a new `routedns.cache.<id>.save-skipped` expvar, so
silent data loss can be graphed and alerted on rather than only grepped for. It sits
alongside the existing `hit`/`miss`/`entries` metrics under the same cache id, via a new
optional `MemoryBackendOptions.MetricsID` (defaulting to `memory` when unset, so the
exported constructor stays source-compatible). The Redis backend has an analogous
`async-skipped` counter, though that one is not per-instance.

## Benchmarks

1000 A-record answers, `-benchmem`, best-of-6 on an i5-7200U. A Cortex-A53 is
considerably slower in absolute terms, but the ratios should hold.

| | before | after | change |
|---|---|---|---|
| Serialize | 4.87 ms | 1.07 ms | **~4.5× faster** |
| Serialize allocs | 5012 | 20 | **−99.6%** |
| Serialize bytes | 1.30 MB | 0.79 MB | **−39%** |
| Deserialize | 9.88 ms | 9.92 ms | ~unchanged |

Deserialize is essentially unchanged, as expected — the decode path is untouched. The
buffering win there is in syscall count, which an in-memory benchmark doesn't capture.
It is now the slower half of persistence by ~9×; converting it to a matching hand-rolled
decoder is the obvious follow-up, but it needs its own format-compatibility tests and is
left out of this change.

Benchmarks (`BenchmarkLRUSerialize` / `BenchmarkLRUDeserialize`) are included; the
before-numbers were taken by running the same benchmark file against the unmodified
implementation.

## Tests

New `lru-cache-serialize_test.go`:

- `TestLRUSerializeMatchesEncodingJSON` — output is byte-identical to `encoding/json`,
  covering plain records, every key field populated (ECS/DO/CD), and names requiring
  JSON escaping (quotes, non-ASCII, control characters, and the `<` `>` `&` that
  `encoding/json` escapes by default). It builds its expectation by running
  `encoding/json` at test time rather than against a golden string, so adding or
  reordering a field in `lruKey`, `cacheAnswer` or `dns.Question` fails it — that is
  what keeps the hand-rolled encoder from silently drifting from the struct
  definitions. Note `dns.Question` belongs to miekg/dns, so a dependency bump can
  break it too; `appendCacheItemJSON` carries a comment saying so.
- `TestLRUSerializeRoundTrip` — keys, messages and timestamps survive a save/load.
- `TestLRUSerializePreservesLRUOrder` — LRU ordering survives persistence, so the right
  entries are evicted when loading into a capacity-limited cache.
- `TestLRUSerializeSkipsExpired` / `TestLRUDeserializeSkipsExpired`.
- `TestLRUDeserializeOldFormat` — a file written by the previous encoder still loads.
- `TestLRUSerializeSkipsUnpackableMessage` — one bad record doesn't fail the save, and
  is reported to the caller in the returned count.
- `TestMemoryBackendFileRoundTrip` — covers the real `writeToFile`/`loadFromFile` path
  including the buffered writer's flush, which had no test coverage before.
- `TestLRUSerializeSkipsAreLoggedOnce` — unpackable records produce one summary log
  line with a count, not one per record.
- `TestMemoryBackendSaveSkippedMetric` — the skip count reaches the expvar, not just
  the return value.
- `TestCacheAnswerLive` — pins the shared expiry boundary: a record is live up to and
  including its expiry instant.
- `TestMemoryBackendConcurrentSaves` — overlapping saves of differing lengths must
  leave a complete, parseable file and no stray temp files. This passes with `saveMu`
  removed (the rename is what guarantees integrity); what it catches is a regression
  to writing the target file directly.

New `atomicfile_test.go` covers the shared helper directly, including the error path
that is the whole reason it exists — a callback that fails part-way must leave the
previous file byte-intact and clean up its temp file.

New `blocklistloader-http_test.go` covers `writeToDisk`/`loadFromDisk`, which had no
test file at all. This matters because the helper derives the temp directory from the
target filename rather than from `opt.CacheDir` as the old code did; the tests assert
the list round-trips, lands at the hashed cache filename, and leaves nothing else in
the cache-dir. Both new test files were checked by mutation — reverting
`writeFileAtomic` to a direct `os.Create` makes them fail.

The "no temp file left behind" check every `writeFileAtomic` caller wants is a single
`requireOnlyFile` helper rather than five copies, and it also asserts the surviving
file is the intended target — which three of the original call sites didn't check.

Full suite passes, including under `-race`.

Note the `-race` run was previously red: `TestLRUSerializeSkipsAreLoggedOnce` swapped
the package-global `Log` to capture output, racing against background goroutines from
other tests that read it unsynchronised (`client-blocklist.go`). This reproduces at
the parent commit, so it isn't introduced here, but it is fixed: `serialize` now takes
the logger to write to, so no test mutates the global.

## Notes

While testing I hit a pre-existing issue, **not addressed here**: `Lookup` computes
`age := uint32(time.Since(timestamp).Seconds())` (`cache-memory.go`, and the same in
`cache-redis.go`). If `Timestamp` is in the future — a clock step backwards, or a cache
file carried between machines — the conversion wraps to ~4 billion and every record
looks expired, silently emptying the cache. Happy to raise it separately.


lru-cache.go
```diff
@@ -190,17 +204,132 @@ func (c *lruCache) size() int {
...
+//
+// Records are encoded by appending to a reused buffer rather than via
+// json.Encoder, which would reflect over every record and re-parse the
+// output of cacheAnswer.MarshalJSON to compact it.
+//
+// Records that can't be packed are dropped rather than failing the whole
+// save, and summarised in a single log line so a systematic problem doesn't
+// turn into one line per record on every save-interval. Returns how many
+// were dropped, for the caller to surface as a metric.
+func (c *lruCache) serialize(w io.Writer, log *slog.Logger) (skipped int, err error) {
```
folbricht: Don't like the new logger argument here.

A lot of effort went into supporting "skipped". We have a whole new metric, passing around loggers, etc, all to support something that isn't really needed. The message packing shouldn't fail, and if it does we can decide to error or warn+ignore. We shouldn't need to count those failures.

lru-cache.go
```diff
@@ -190,17 +204,132 @@ func (c *lruCache) size() int {
...
+}
+
+// Appends one cache item to buf as a single JSON line, matching the layout
+// produced by encoding/json for cacheItem/cacheAnswer.
+//
+// The field list below must track lruKey, cacheAnswer, and dns.Question --
+// the last owned by miekg/dns, so a dependency bump can also break this.
+// TestLRUSerializeMatchesEncodingJSON compares against encoding/json at test
+// time and fails if any of the three gains or reorders a field.
+func appendCacheItemJSON(buf []byte, item *cacheItem, wire []byte) []byte {
```
folbricht: Not a fan of this low-level marshaling of json here. Yes it's probably faster that the (current) std lib, but at the cost of complexity. I think it may be worth waiting for https://pkg.go.dev/encoding/json/v2 which should come in the next Go version (1.27.x) to see if that'll give us the performance without the extra complexity.

lru-cache.go
```diff
@@ -211,6 +340,12 @@ func (c *lruCache) deserialize(r io.Reader) error {
 		if item.Key.Question.Name == "" || item.Answer == nil {
 			continue
 		}
+		// Skip records that expired while the file sat on disk; until the
+		// next lookup or GC run they'd occupy capacity live records could use.
+		// Files written by older versions aren't filtered on the write side.
+		if !item.Answer.live(now) {
```
folbricht: This makes sense, but do we really need a helper function for something so trivial that's only needed in a couple of places? I'd remove the `.live()` helper, it's just unnecessary.

atomicfile.go
```diff
@@ -0,0 +1,51 @@
...
+const fileBufSize = 64 * 1024
+
+// Replaces the file at filename with whatever write produces, atomically.
+// The content goes to a temporary file in the same directory which is then
+// renamed into place, so an error part-way through leaves the previous file
+// intact rather than a truncated one. The writer passed to write is buffered;
+// it's flushed and synced before the rename.
+//
+// The temp file is fsync'ed before the rename, so the rename can't be made
+// durable ahead of the data it points at -- without that, a machine crash
```
folbricht: Yes, but the fix creates an entirely new problem where a crash could leave the old temp file behind. Each os.CreateTemp() creates a new file with a unique name that could stay on disk.

Perhaps instead we don't use os.CreateTemp() but just a regular os.Create() with a name derived from the target, so after a crash the next write would clean it up.

atomicfile_test.go
```diff
@@ -0,0 +1,72 @@
+package rdns
```
folbricht, line 1: There's way too many tests for a simple function like `writeFileAtomic`, wonder if we need any at all.

folbricht (commented):
Optimizing some of these makes sense for sure. But let's first break this PR up into:
1. That handles the file writes
2. Something that deals with the file format

The first is pretty straightforward. As for the 2nd part, perhaps this could be solved in two independent phases.

2a. To improve the JSON marshaling performance, most of the gains can be achieved by removing the double-encode with MarshalJSON and instead using a single write, plus waiting for json/v2 when that becomes the default.

2b. If performance is so critical, we probably shouldn't be using JSON on constrained devices and offer another file format (in addition to json), something that can be written fast and efficiently.

emlimap:
@folbricht Thanks — split done. This PR now covers only the file writes; the format work is out and will come separately.

Addressed:

- Reverted the append encoder and `appendCacheItemJSON`. `lru-cache.go` is back to exactly what's on master.
- Dropped the skipped-record count, the `save-skipped` expvar and the logger argument to `serialize()`. Agreed it wasn't paying for itself.
- Dropped the `.live()` helper.
- Cut the `writeFileAtomic` tests right down.

## One deviation, on the temp file

I tried the derived name and had to back it out — it introduces file corruption.

With a fixed `<target>.tmp`, two writers to the same file both open it, write at independent offsets into one inode, and then both rename it into place. The result isn't a lost update, it's a file containing a mix of both writers' output, which the rename then publishes. I measured 17 corrupt results in 100 runs.

It only shows up once the content exceeds the 64KiB write buffer, so a small file never trips it — but a blocklist always does.

It's reachable on the blocklist path: `HTTPLoader` has no lock, `cacheFilename()` is `sha256(url)` under the cache-dir, and each list gets its own refresh goroutine on its own timer. Two lists pointing at the same URL sharing a cache-dir — or the same list used as both a blocklist and an allowlist — write concurrently. `loadFromDisk` doesn't validate, so a corrupt cache is loaded verbatim at the next startup.

`os.CreateTemp`'s unique names are what prevent this, so I've kept them and dealt with the orphans directly instead: `removeStaleTempFiles` sweeps `<target>.tmp*` when the backend or loader is constructed. That gets the property you wanted — leftovers don't accumulate across restarts — without trading correctness for it.

Sweeping at startup rather than before each write is deliberate. A leftover can only come from a previous run, so a per-write sweep would re-read the whole directory on every save to clean up something that can't have appeared since the process started, and a shared cache-dir is exactly the case where that directory is large. It would also be racy: the glob would match the temp file of a write already in flight.

`TestWriteFileAtomicConcurrent` covers the corruption case; `TestRemoveStaleTempFiles` covers the cleanup.

Happy to change the approach if you'd rather handle the leftovers another way.

## Two smaller things

- The callback takes `io.Writer` rather than `*bufio.Writer`, so the buffering stays an implementation detail.
- Writing blocklist rules no longer allocates a string per rule.

## Worth flagging

Blocklist cache files change mode from 0600 to 0644. `os.CreateTemp` creates at 0600 and `rename` preserves the mode, whereas the old code wrote the target directly with `os.Create`. The DNS cache file was already 0644 so it's unchanged there, and blocklist content is public, but it's a change and I didn't want it to pass silently. Easy to pin back to 0600 with an explicit `os.Chmod` before the rename if you'd prefer.

atomicfile.go
```diff
@@ -0,0 +1,74 @@
...
+// truncated one. The writer passed to write is buffered and flushed before the
+// rename.
+//
+// The temp file gets a unique name rather than one derived from the target.
+// Two writers to the same file would otherwise share one temp file, writing
+// into it at independent offsets and renaming the result into place: not a
+// lost update but a corrupt mix of both. Any leftover from a crash between
+// writing and renaming is cleaned up by removeStaleTempFiles at startup.
+func writeFileAtomic(filename string, write func(w io.Writer) error) error {
+	tmp, err := os.CreateTemp(filepath.Dir(filename), filepath.Base(filename)+tmpSuffix)
```
folbricht: Cache file permissions change from 0644 to 0600 here. `memoryBackend.writeToFile` previously used `os.Create` (0666 & ^umask, so typically `-rw-r--r--`), whereas `os.CreateTemp` hardcodes 0600 and `os.Rename` carries that mode onto the target.

After upgrading, the first save replaces an existing world-readable `/var/tmp/cache.json` (the path shown in `doc/configuration.md`) with a `-rw-------` file, which breaks any external reader or monitoring, and an admin-set mode gets reset on every save. A `tmp.Chmod` to 0644, or to the existing target's mode when there is one, would keep the previous behaviour. The blocklist path is unaffected, it already used `CreateTemp`.

atomicfile.go
```diff
@@ -0,0 +1,74 @@
...
+	// Flush to disk before the rename, so a machine crash can't leave the new
+	// name pointing at an empty file.
+	if err := tmp.Sync(); err != nil {
+		return err
+	}
+	// Close before renaming; Windows needs it.
+	if err := tmp.Close(); err != nil {
+		return err
+	}
+	return os.Rename(tmpName, filename)
```
folbricht, line 138: `os.Rename` does not follow a symlink at the destination, so a symlinked target is now replaced rather than written through, where `os.Create` used to follow it.

On the OpenWRT style setups this PR targets, `/etc/routedns/cache.json` symlinked to `/tmp/cache.json` is a common way to keep a stable config path while the data lives on tmpfs. With this change the first save unlinks the symlink and leaves a regular file in `/etc`, so the file lands on flash and the tmpfs target is orphaned. Probably worth a note in the docs at least.

Security-wise the change is an improvement, since it removes a symlink-attack vector in the world-writable `/var/tmp` the docs suggest.

atomicfile.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Removes temp files left next to filename by a writeFileAtomic call that was
+// interrupted before it could rename or clean up. Errors are ignored; a
+// leftover file wastes space but doesn't affect correctness.
+//
+// Call this at startup, not before each write: a leftover can only come from a
+// previous run, and the glob would otherwise match, and delete, the temp file
+// of a write that's currently in flight.
+func removeStaleTempFiles(filename string) {
+	matches, err := filepath.Glob(filename + tmpSuffix + "*")
```
folbricht: `filename` comes straight from config (`cache.backend.filename`, `cache-dir`) and is used here as a glob pattern without escaping.

A `[` in the path makes `Glob` return `ErrBadPattern`, so the sweep silently does nothing and leftovers accumulate forever. A `*` or `?` lets the pattern reach past its own target: `filename = /var/cache/c*.json` produces the pattern `/var/cache/c*.json.tmp*`, which can match and delete another cache's in-flight temp file in the same directory.

Escaping the literal prefix, or reading the directory and matching with `strings.HasPrefix`, avoids both.

folbricht:
Following up on the temp-file comments above, here is a concrete suggestion that would address them together.

The sweep is currently keyed on the target filename, which leaves a class of orphans it can never see. If a blocklist source URL changes or a list is dropped from the config, its `<sha256>.tmp*` leftover no longer has a live target to be swept from, so it stays forever. Same for the `routedns*` temps written by the previous `writeToDisk`. On a device where the cache-dir is a small tmpfs and a hosts list is tens of MB, one stranded temp file is significant.

Sweeping the directory by prefix and age, rather than globbing the target name, handles all of that:

```go
// Marks a temp file written by writeFileAtomic. Deliberately not derived from
// the target name: a sweep keyed on the target can't reclaim temps whose target
// has since been removed from the config.
const tmpPrefix = ".routedns-tmp-"

// Temps younger than this belong to a write that may still be in flight,
// possibly in another process sharing the directory.
const tmpMaxAge = time.Hour

func removeStaleTempFiles(dir string) {
	entries, err := os.ReadDir(dir)
	if err != nil {
		return
	}
	cutoff := time.Now().Add(-tmpMaxAge)
	for _, e := range entries {
		if !strings.HasPrefix(e.Name(), tmpPrefix) {
			continue
		}
		info, err := e.Info()
		if err != nil || info.ModTime().After(cutoff) {
			continue
		}
		os.Remove(filepath.Join(dir, e.Name()))
	}
}
```

with `writeFileAtomic` using `os.CreateTemp(filepath.Dir(filename), tmpPrefix)`.

What that buys:

- Orphans whose target is gone from the config get reclaimed, because the sweep no longer needs to know the target names.
- No glob, so a config path containing `[` no longer turns the sweep into a silent no-op, and one containing `*` or `?` can no longer reach sideways into another cache's in-flight temp.
- The age gate makes the sweep safe to run while another writer is live, which matters because the docs suggest a shared `cache-dir = "/var/tmp"`. An in-flight temp is by definition freshly modified. This removes the need for the "a leftover can only come from a previous run" assumption, which doesn't hold across processes.
- Since it is then safe alongside a live writer, it can also run on a slow timer instead of only at construction, so a box that stays up for months doesn't accumulate. Hourly, not per save; the objection to re-reading the directory on every write still stands.
- Adding `routedns` to the prefix check as a one-time migration reclaims the pre-upgrade leftovers.

Two things worth keeping exactly as they are, since both look like tempting simplifications:

**The temp has to stay in the target's directory.** Moving it to the system temp dir wouldn't merely weaken the atomicity, it would break the write outright: Go's `os.Rename` has no copy fallback and returns `EXDEV` across filesystems, and `/tmp` is tmpfs on the devices this targets.

**The unique random name is load-bearing beyond the interleaving corruption** described in the commit message. `os.CreateTemp` opens with `O_CREATE|O_EXCL` and retries on collision, so it refuses to follow a symlink planted at that path. A fixed `<target>.tmp` would have to use `os.Create` (`O_CREATE|O_TRUNC`, no `O_EXCL`), which does follow one. Blocklist target names are `sha256(url)` and so predictable from the config, which in a world-writable cache-dir would let someone pre-plant `<sha256>.tmp` as a symlink and have an arbitrary file truncated and overwritten with list content on the next refresh. Probably worth a comment saying so, so the orphan problem doesn't later get "fixed" by making the name deterministic.

The `os.Remove` in the sweep is safe against that same trick, for what it's worth: `unlink` operates on the link rather than its target, and `ReadDir`'s `Info()` is an `lstat`, so the age check doesn't follow links either.

folbricht:
@emlimap Side-note: From next weekend I'll be adding improvements to help run routedns on resource-constrained devices, like OpenWRT. There are a lot of low-hanging fruit in terms of CPU/memory usage and I'll be hacking away at them (already started with some logging fixes). A faster file format for storing caches will be one of them. Feel free to open issues for any suggestions

emlimap:
@folbricht if this is causing too many back and forth and makes it easier for you to make the change yourself, feel free to. Don't want to cause too much burden on you by using AI.

---
Thanks, all four are fixed, and the directory-and-age design is what's in there now.

Permissions. The temp takes the mode of the file it replaces, or 0644 when there isn't one. I also made a stat that fails for any reason other than not-exists an error rather than a fallback — a symlink loop or an unsearchable parent directory would otherwise have quietly turned a 0600 cache into a 0644 one, which is the wrong direction for a file that records what's been looked up.

One behaviour difference worth flagging rather than leaving to be discovered: Chmod isn't masked by umask, where os.Create was. Under umask 0077 the old code produced a 0600 cache file and this produces 0644. That's the intended behaviour for the case you raised, since an existing world-readable file keeps its mode, but it does mean an operator's umask no longer applies to a new cache file. Happy to mask it if you'd rather umask kept winning there.

Glob injection. Both cases were real — I checked before changing anything. A [ gives ErrBadPattern so the sweep silently does nothing, and a cache-*.json target produces a pattern that matches another cache's in-flight temp in the same directory. There's no glob now, just a ReadDir and a prefix check.

The redesign. Implemented as you described: .routedns-tmp- prefix, one hour, os.CreateTemp(filepath.Dir(filename), tmpPrefix). Two adjustments on top:

The legacy routedns prefix is matched only when followed by the digits os.CreateTemp appends. A bare prefix match would take oog, routedns.pid and the binaryitself, and cache-dir = "/var/tmere those might sit. The sweepalso skips anything that isn't a regular file, so a directory that happens to match the pattern isn't removed.

filepath.Dir returns "." for a bare filename = "cache.json", so that and the empty string are skipped. Without that, a config using a relative filename would sweep the working directory — which is where the binary and the config file usually are. The trade-off is that temps in
that case accumulate rather thano sweep seemed clearly betterthan the alternative.

Your point about the age gate hoch flush, so an in-flight write
keeps looking fresh. The one gapover an hour with less than 64KiB buffered, since nothing has reached the file yet. If that does happen the failure is safe
— the writer's fd stays valid, t and the previous cache file
survives.

Symlinks. Documented on the filename option. Both points confirmed: os.Rename replaces the symlink with a regular file, and O_EXCL refuses to follow a symlink planted at the temp path
where os.Create would truncate weasoning is now a comment abovewriteFileAtomic so the random name doesn't get "simplified" away later.

Tests are four in that file, covering the write path, the concurrent case, the sweep's keep/remove decisions, and the a one fails if I revert the fix it covers.

folbricht:
Two things from a read through this branch.

### New files bypass the umask and end up world-readable

In `writeFileAtomic`, `os.CreateTemp` gives us a 0600 file and the `tmp.Chmod(mode)` then sets it to `defaultFileMode` (0644) whenever the target doesn't exist yet. `chmod(2)` isn't filtered by the process umask, unlike the create-mode argument that `os.Create` used to pass, so the "Matches what os.Create would have produced" comment only holds at umask 022.

Under `umask 077`:

```
os.Create        -> -rw-------
writeFileAtomic  -> -rw-r--r--
```

Two concrete regressions:

* A memory cache under systemd `UMask=0077`, or any similarly hardened deployment, used to be 0600 and is now 0644. That file is a record of every domain the network resolved, which is the concern the comment above the `os.Stat` block already raises.
* The blocklist cache file was previously `os.CreateTemp` plus rename with no chmod at all, so it was 0600 unconditionally regardless of umask. It is now 0644, in the `cache-dir = "/var/tmp"` that `doc/configuration.md` recommends.

This only affects files that don't exist yet, since existing ones keep their mode through the `os.Stat` branch, so it shows up gradually rather than all at once.

Either skipping the `Chmod` when the target is new, which leaves `os.CreateTemp`'s 0600 and matches what the blocklist path did before, or masking `defaultFileMode` with the process umask, would fix it.

### Suggest dropping `legacyTmpPrefix`

`isTempFileName` accepts `routedns` followed by any run of decimal digits. The comment is right that this protects `routedns.toml` and a bare `routedns` binary, but it also matches version and date stamped neighbours:

```
routedns          -> false
routedns.toml     -> false
routedns2         -> true
routedns0231      -> true
routedns20250101  -> true
```

`removeStaleTempFiles` now runs on `filepath.Dir(opt.Filename)` and on `opt.CacheDir`, which are directories the user picked for one particular file rather than directories routedns owns, and routedns generally runs as root since it binds :53, so the sticky bit on `/var/tmp` doesn't prevent unlinking someone else's matching file. An operator keeping a pinned `/var/lib/routedns/routedns0231` next to `filename = "/var/lib/routedns/cache.json"` would lose it on the next start.

Tightening the digit run to a fixed width doesn't help: `os.CreateTemp` formats a random `uint32` with no padding, so it emits anywhere from 1 to 10 digits and `routedns2` is a name it could have produced itself.

What the legacy match buys looks smaller than that risk. Only `blocklistloader-http.go` ever created temp files before this change, its deferred `os.Remove` already covers every normal error path, so a leftover requires a hard kill in the window between `CreateTemp` and that defer, and each one is a single blocklist copy that gets re-downloaded anyway. The memory cache used plain `os.Create` and never wrote a temp file, so in `filepath.Dir(opt.Filename)` there is nothing legacy to find in the first place and the digit match is pure downside there.

Dropping it costs at most one stale file per historical crash in one directory, and removes the whole false positive class.

emlimap:
Both correct, both fixed.

Permissions. Reproduced before changing anything:

umask 0022: os.Create=-rw-r--r--  writeFileAtomic=-rw-r--r--
umask 0077: os.Create=-rw-------  writeFileAtomic=-rw-r--r--

I'd flagged the chmod/umask difference earlier as a behaviour change worth noting, but I had the trade-off backwards — I treated "matches os.Create at the default umask" as the thing to preserve, when the point of setting a umask is that it applies. And your second case is worse than the first: the blocklist path was os.CreateTemp + rename with no chmod at all, so it was 0600 at any umask, and cache-dir = "/var/tmp" is what the docs recommend.

Fixed by skipping the Chmod when the target doesn't exist, so a new file keeps os.CreateTemp's 0600.
I went that way rather than maskiading the process umask turns outnot to be worth it. syscall.Umaskading means setting it twice, andumask is per-process — that window widens any concurrent file creation anywhere in the processWith four goroutines calling os.C under umask 077, somewherebetween a third and a half of the 8000 files came out world-readable depending on scheduling. That would be adding a permissionug, in a process that also writesthe query log. syscall.Umask doesn't compile on Windows either, which this package builds for, and /proc/self/status is Linux-only and not there in the distroless image.

Skipping the chmod needs no platform code and no build tags, and it matches the usual pattern — lumberjack does exactly this for rotated logs: mode := os.FileMode(0600), then copy the mode off the existing file if there is one, and never look at umask. Neither does the stdlib; it leaves that to open(2).

0600 also seems right for the content rather than just conservative: the memory cache file is every domain looked up with timestamps and answers, plus the client subnet when ECS is in play (lru-cache.go:230). And both shipped units set DynamicUser=true, so the uid isn't stable across
restarts and a world-readable cac either. An existing file stillkeeps its mode, so the upgrade case you raised originally is unaffected — only new files change.

Documented on both filename and the blocklist cache-dir, since cache-dir is the one whose behaviour actually moved for existing users.

Legacy prefix. Dropped. Your table is what it did:

routedns2         -> true
routedns0231      -> true
routedns20250101  -> true

And you're right that no width reom is strconv.FormatUint of arandom uint32 with no padding, so 1 to 10 digits, and routedns2 is a name it could have produced. I'd checked that the match couldn't reach routedns.toml or the binary and stopped
there, without asking what else i

The cost of dropping it is smaller than I'd assumed, too. On master the memory cache used plain
os.Create and never wrote a temp name) there was never anythinglegacy to find — the match was pure downside in that directory. Only the blocklist loader wrote temps, its deferred os.Remove covers the normal error paths, and each leftover is one blocklist copy that gets re-downloaded.

isTempFileName became a one-line prefix check with one caller, so it's inlined and the reasoning moved to a comment on the sweep.

Net −13 lines. Tests stay at four/ routedns0231 / routedns20250101names moved into the existing bystander list rather than adding a case, and the new-file assertion now pins 0600. I checked both fixes fail the suite when reverted, and that the tests pass identically under umask 022 and umask 077.

---
Everything factual in it is verifred. Two things I trimmed whiledrafting: a claim that lumberjack was already in this module graph (it isn't — it was only in my local module cache, so it's now cnd a renameio citation I couldn't confirm the details of.

folbricht (approved):

folbricht:
Thank you

Issue #609, opened 2026-08-09
Optional build tags to reduce binary size on flash-constrained devices


emlimap:
On resource constrained devices, not only memory and CPU are limited, storage is too. There are OpenWrt devices with 32 or 64 MB of storage where any savings would be meaningful.

I was thinking we should explore supporting build tags to selectively build only the features that are used by the individual. I'm unsure how much effort this would require or how feasible it is given the project's current state, but I wanted to start a discussion. I asked Claude to investigate, and here's what came back.

The likelihood of people using Redis as a memory backend while running RouteDNS on OpenWrt is quite low, so an option to build without it can be beneficial. Maybe an additional release artifact can be beneficial as well without having to compile it if isn’t much work

---
Measured on Go 1.26.5, linux/amd64, `-trimpath -ldflags="-s -w"` to match the release build. Baseline **21.35 MiB**, which matches the v0.1.232 linux-amd64 release (21.02 MiB) closely enough to be representative.

## Where the binary goes

Third-party text, largest first:

| Library | KiB |
| --- | ---: |
| `redis/go-redis` | 2751 |
| `quic-go/quic-go` | 652 |
| `folbricht/routedns` | 449 |
| `miekg/dns` | 443 |
| `yuin/gopher-lua` | 343 |
| `pion/dtls` | 236 |
| everything else | < 500 combined |

The project's own code is ~450 KiB of a 21 MiB binary; it's essentially all dependencies.

## What removing things actually saves

Each measured by stripping the feature and rebuilding:

| Removed | Binary | Saving |
| --- | ---: | ---: |
| redis backend | 16.23 MiB | **−5.1 MiB (24%)** |
| lua | 20.37 MiB | −1.0 MiB (5%) |
| DTLS | 20.62 MiB | −0.73 MiB |
| ODoH | 20.99 MiB | −0.36 MiB |
| DoQ | 21.26 MiB | −0.09 MiB |
| DoQ + DTLS + ODoH | 20.23 MiB | −1.12 MiB |

redis is worth more than everything else combined. It's also larger than its text size suggests — 2.75 MiB of text but 5.1 MiB off the binary, because it brings its own transitive tree.

## Three things that surprised me

<details>
<summary><b>Stripping DoQ saves nothing</b> — DoH keeps quic-go linked</summary>

[...]

PR #615, opened 2026-08-15
Add build tags to leave out redis and lua

folbricht:
Addresses #609.

The binary is almost entirely dependencies, and the two largest optional ones are the Redis client and the Lua interpreter. Two build tags leave them out:

| Tag | Leaves out |
| --- | --- |
| `noredis` | The `redis` cache backend |
| `nolua` | `lua` groups |

Sizes with `-trimpath -ldflags="-s -w"`, as the release builds use:

| Target | Current | Both tags |
| --- | ---: | ---: |
| linux/mipsle, softfloat | 23.5MB | 17.0MB |
| linux/armv6 | 20.6MB | 14.9MB |
| linux/arm64 | 20.4MB | 14.6MB |
| linux/amd64 | 22.2MB | 15.9MB |

Redis accounts for most of it. It contributes 2.75MB of text but takes 5.4MB off the binary because of its transitive tree. Lua is worth about 1MB.

One caveat on what this buys in practice: OpenWrt stores its root filesystem on squashfs and package payloads are compressed, so the number that matters there is the compressed size. For linux/mipsle that goes from 4.51MB to 3.74MB under xz, a 17% saving rather than the 28% the raw sizes suggest. Still useful on a 16MB device, but less dramatic.

## Approach

Configuration is unchanged in both builds. The config layer parses every option either way, and the stubs replacing the two features return an error naming the missing one, so a config asking for a feature the binary lacks fails at startup instead of being silently ignored:

```
$ routedns-minimal config.toml
Error: this build does not support the redis cache backend
```

`RedisBackendOptions` used to embed `redis.Options`, which meant `cmd/routedns` imported the client and kept it linked no matter what the library did. Its connection settings are now plain fields that the backend translates, and `NewRedisBackend` returns a `CacheBackend` and an error rather than the unexported `*redisBackend`. Both option structs live in untagged files so there is only one definition of each.

`TestBackendsAgeIdentically` moved from `cache-age_test.go` to `cache-redis_test.go`, since it goes through the Redis codec and cannot build without it. The rest of `cache-age_test.go` is backend independent and stays where it is.

## Release artifacts

A `minimal` goreleaser build produces `routedns-minimal-linux-{armv6,armv7,arm64,mips,mipsle}`. Existing artifact names and the deb/rpm/apk/archlinux packages are untouched. Verified with `goreleaser check` and a full `release --snapshot`.

CI builds and runs the test suite with both tags set, otherwise the stubs would drift out of sync unnoticed.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -26,3 +26,11 @@ jobs:
 
     - name: Test
       run: go test -race ./...
+
+    # The reduced release builds replace the Redis backend and the Lua
+    # resolver with stubs. Nothing else exercises those, so build and test
+    # them here or they break unnoticed.
+    - name: Build and test without optional features
+      run: |
+        go build -v -tags nolua,noredis ./...
+        go test -tags nolua,noredis ./...
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -70,12 +70,52 @@ builds:
     gomips:
       - softfloat
 
+  # Same small-device targets again, built without the Redis cache backend
+  # and without Lua. Those two dependencies are about a third of the binary,
+  # and neither is likely to be used on a router with 16MB of flash. The
+  # config options remain, a config that uses them fails at startup with a
+  # message saying the build doesn't support them.
+  - id: minimal
+    binary: routedns
+    main: ./cmd/routedns/
+    env:
+      - CGO_ENABLED=0
+    flags:
+      - -trimpath
+    tags:
+      - nolua
+      - noredis
+    ldflags:
+      - -s -w
+    goos:
+      - linux
+    goarch:
+      - arm
+      - arm64
+      - mips
+      - mipsle
+    goarm:
+      - "6"
+      - "7"
+    gomips:
+      - softfloat
+
 # Upload plain binaries named like the previous handcrafted releases
 # (routedns-linux-amd64, routedns-linux-armv7, ...) so existing download
-# links keep working.
+# links keep working. The reduced builds carry a -minimal suffix.
 archives:
-  - formats: [binary]
+  - id: default
+    ids:
+      - packaged
+      - embedded
+    formats: [binary]
     name_template: '{{ .ProjectName }}-{{ .Os }}-{{ .Arch }}{{ with .Arm }}v{{ . }}{{ end }}'
+
+  - id: minimal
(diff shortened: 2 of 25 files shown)


emlimap:
@folbricht Finally got some time to test this change today. Thank you for opening this PR

This change was tested on Nano Pi R3S (ARM64) with 1GB of RAM running OpenWrt 25.12.5. I can confirm that the binary size has dropped. I also went ahead and tested with just `noredis` option as I have started experimenting with Lua, so I might keep it.

Something to note is that routedns is running under ujail sandbox (https://openwrt.org/docs/guide-developer/procd-init-scripts#service_jails) so the numbers might slightly differ to someone running it without sandboxing. Also have this env var set
```sh
GOMEMLIMIT=128MiB
```

Below are the test results. I ran it with the help of Claude to produce a repeatable test.


**Binary size** (`-trimpath -ldflags="-s -w"`):

| | Full | `noredis` | `noredis,nolua` (minimal) |
|---|---:|---:|---:|
| Size | 19.5 MB | 15.0 MB | 14.1 MB |
| vs. full | — | -23.1% | -27.7% |

Matches the PR description's arm64 numbers closely, and confirms redis is the larger of the two contributors.

**Startup RSS**, median of 5 restarts each, sampled via `ps -o rss=` after one warm-up query:

| | Full | `noredis` | `noredis,nolua` (minimal) |
|---|---:|---:|---:|
| Median | 19764 KB | 18972 KB | 18300 KB |
| vs. full | — | -4.0% | -7.4% |
| Range (5 samples) | 19076–21788 KB | 18912–23164 KB* | 16232–18336 KB |

\* one outlier (23164) in the `noredis` run; excluding it the other 4 samples cluster tightly around 18.9–19.0 MB.

RSS is noisy on this hardware (multi-MB spread run-to-run for the *same* binary), but both reduced builds trend consistently at or below the full build across samples — a modest, repeatable improvement, not pure noise. No RSS regression in any configuration.

**Functional check**: both reduced binaries correctly reject a config that references the disabled feature (`redis` cache backend / `lua` groups) at startup with the documented error, rather than silently ignoring them. Normal DNS resolution (UDP, cache) works identically across all three binaries.

Ran under the actual `procd`/ujail service wrapper (not a bare foreground process), so this reflects real deployment conditions, not just a manual test run.

PR #651, opened 2026-09-11
Notify readiness to systemd

bbjubjub2494:
On Linux, systemd supports an IPC protocol that allows daemons to signal to PID 1 that they have finished starting up. (See [`sd_notify(3)`](https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html)) This PR uses a pure-Go implementation to bring this functionality to routedns.

This is useful because this way, systemd units that specify `After=routedns.service` will be guaranteed that routedns is listening on its sockets. Otherwise there is a race condition which I ran into. (In a slightly unusual set up, i.e. I run a NixOS deployment and many systemd units including routedns and wireguard setup units that need DNS resolution restart in the same transaction)

The code is no-op in existing setups that don't explicitly enable the feature.

### DIFF
--- a/cmd/routedns/main.go
+++ b/cmd/routedns/main.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	syslog "github.com/RackSec/srslog"
+	"github.com/coreos/go-systemd/v22/daemon"
 	rdns "github.com/folbricht/routedns"
 	"github.com/heimdalr/dag"
 	"github.com/redis/go-redis/v9"
@@ -549,6 +550,14 @@ func run(opt options, args []string) error {
 			continue
 		}
 		go superviseNetNSListener(pl.id, pl.nsName, pl.build)
+	}
+
+	// Notify Systemd that we are done starting, i.e. the listeners are up.
+	// This is a no-op unless $NOTIFY_SOCKET is set.
+	_, err = daemon.SdNotify(false, daemon.SdNotifyReady)
+	if err != nil {
+		rdns.Log.Error("daemon notification failed", "error", err)
+		// carry on
 	}
 
 	// Graceful shutdown
--- a/cmd/routedns/routedns.service
+++ b/cmd/routedns/routedns.service
@@ -3,7 +3,7 @@ Description=RouteDNS - DNS stub resolver and router
 After=network.target
 
 [Service]
-Type=simple
+Type=notify
 DynamicUser=true
 NoNewPrivileges=true
 WorkingDirectory=/opt/routedns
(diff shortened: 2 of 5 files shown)


folbricht:
The notify call lands after the loop that starts the listener goroutines, so
READY=1 is sent once the process is up rather than once the listeners are
accepting. Nothing in `build()` binds a socket: every listener type does its
`net.Listen`/`ListenAndServe` inside `Start()`, which at that point is running
concurrently, and `daemon.SdNotify` is a `net.DialUnix` plus a `Write`, so it
usually completes before those goroutines are scheduled. A unit ordered
`After=routedns.service` can still hit connection refused, and a listener that
can never bind (address in use, missing `CAP_NET_BIND_SERVICE`) leaves the retry
loop logging while systemd has already been told the service is ready.

That is still a clear improvement over `Type=simple`, where readiness is reached
at exec, before config parsing and blocklist loading. Those dominate startup, so
the remaining window is short. Worth tightening in a follow up: `NotifyStartedFunc`
on `dns.Server` covers the dns, dot and dtls listeners, while doh, doq and odoh
would need `Start()` split into a bind step and a serve step, with the notify sent
once all of them report in. The netns supervised listeners are gated on namespace
presence and cannot join that set, so they would stay outside readiness.

Suggest adjusting the comment to say the service is up rather than the listeners.

One build note: the `go-systemd` import is placed after the `rdns` import, but
gofmt sorts by import path, not by alias. `gofmt` is enabled in `.golangci.yml`,
so the lint job will fail until the line moves above `rdns "github.com/folbricht/routedns"`.

bbjubjub2494:
I missed the fact that the listeners are asynchronous and got a bit too confident. I am testing the patch in production so I would have realized that it doesn't fix the race condition eventually, but you saved me some time. My impression is that it still reduces the race condition, but it's not working as intended.

I think the way to implement this properly is to add a `NotifyStartedFunc` callback in our `ListenOptions`, and either pass it through for listener types implemented by `dns.Server`, or call it ourselves before `http.Serve`. Looking into it.

This also means that if some listener is misconfigured and never starts, systemd will eventually time us out. That seems acceptable to me, and there are going to be plenty of logs to figure out why.

folbricht:
I'd be ok with the path as-is since it improves things significantly. The real "fix" is likely more involved so happy to merge and leave that for another PR in the future.

bbjubjub2494:
If you feel confident with this patch, you're welcome to merge it. I'll also take the opportunity to add `notify` to the NixOS systemd unit. I'm also working on #652.

folbricht:
Thank you

PR #652, opened 2026-09-11
Tighten readiness synchronization

bbjubjub2494:
followup to #651

This needs to be tested a bit more. I believe that ODOH listener works without modification because the option is picked up by the underlying DOH listener.

### DIFF
--- a/adminlistener.go
+++ b/adminlistener.go
@@ -98,6 +98,7 @@ func (s *AdminListener) startTCP() error {
 		return err
 	}
 	defer ln.Close()
+	s.opt.notifyStarted()
 	return httpServer.ServeTLS(ln, "", "")
 }
 
--- a/cmd/routedns/main.go
+++ b/cmd/routedns/main.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	syslog "github.com/RackSec/srslog"
+	"github.com/coreos/go-systemd/v22/daemon"
 	rdns "github.com/folbricht/routedns"
 	"github.com/heimdalr/dag"
 	"github.com/redis/go-redis/v9"
@@ -549,6 +550,14 @@ func run(opt options, args []string) error {
 			continue
 		}
 		go superviseNetNSListener(pl.id, pl.nsName, pl.build)
+	}
+
+	// Notify Systemd that we are done starting, i.e. the listeners are up.
+	// This is a no-op unless $NOTIFY_SOCKET is set.
+	_, err = daemon.SdNotify(false, daemon.SdNotifyReady)
+	if err != nil {
+		rdns.Log.Error("daemon notification failed", "error", err)
+		// carry on
 	}
 
 	// Graceful shutdown
--- a/cmd/routedns/main.go
+++ b/cmd/routedns/main.go
@@ -15,6 +15,7 @@ import (
 	"runtime"
 	"slices"
 	"strconv"
+	"sync"
 	"syscall"
 	"time"
 
@@ -358,6 +359,7 @@ func run(opt options, args []string) error {
 
 	// Build the Listeners last as they can point to routers, groups or resolvers directly.
 	var listeners []pendingListener
+	var wg sync.WaitGroup
 	for id, l := range config.Listeners {
 		resolver, ok := resolvers[l.Resolver]
 		// All Listeners should route queries (except the admin service).
@@ -384,9 +386,10 @@ func run(opt options, args []string) error {
 		}
 
 		opt := rdns.ListenOptions{
-			AllowedNet:    allowedNet,
-			NetNS:         netns,
-			SocketOptions: rdns.SocketOptions{FWMark: l.FWMark, BindInterface: l.BindInterface},
+			AllowedNet:        allowedNet,
+			NetNS:             netns,
+			SocketOptions:     rdns.SocketOptions{FWMark: l.FWMark, BindInterface: l.BindInterface},
+			NotifyStartedFunc: wg.Done,
 		}
 
 		var build func() (rdns.Listener, error)
@@ -507,6 +510,7 @@ func run(opt options, args []string) error {
 		}
 		pl := pendingListener{id: id, nsName: nsName, build: build}
 		if pl.nsName == "" {
+			wg.Add(1)
 			pl.ln, err = build()
 			if err != nil {
 				return err
@@ -554,6 +558,7 @@ func run(opt options, args []string) error {
 
 	// Notify Systemd that we are done starting, i.e. the listeners are up.
 	// This is a no-op unless $NOTIFY_SOCKET is set.
+	wg.Wait()
 	_, err = daemon.SdNotify(false, daemon.SdNotifyReady)
 	if err != nil {
 		rdns.Log.Error("daemon notification failed", "error", err)
--- a/dnslistener_test.go
+++ b/dnslistener_test.go
@@ -1,0 +1,155 @@
+package rdns
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+)
+
+type mockNotifier struct {
+	mock.Mock
+}
+
+// Used to mock  a NotifyStartedFunc
+func (m *mockNotifier) NotifyStarted() {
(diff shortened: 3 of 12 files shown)


cmd/routedns/main.go
```diff
@@ -383,9 +386,10 @@ func run(opt options, args []string) error {
 		}
 
 		opt := rdns.ListenOptions{
-			AllowedNet:    allowedNet,
-			NetNS:         netns,
-			SocketOptions: rdns.SocketOptions{FWMark: l.FWMark, BindInterface: l.BindInterface},
+			AllowedNet:        allowedNet,
+			NetNS:             netns,
+			SocketOptions:     rdns.SocketOptions{FWMark: l.FWMark, BindInterface: l.BindInterface},
+			NotifyStartedFunc: wg.Done,
```
folbricht, line 392: `NotifyStartedFunc: wg.Done` is attached to every listener's options, but `wg.Add(1)` below only runs for listeners with `nsName == ""`. A listener with `netns = "foo"` is built and started inside `superviseNetNSListener`, so its successful `Start()` calls `wg.Done()` against a counter that was never incremented for it.

With one udp listener and one netns listener, the netns listener's `Done` can satisfy the udp listener's `Add`, so READY is sent before the udp socket is bound, and the second `Done` panics with `sync: negative WaitGroup counter`. With only netns listeners, `wg.Wait()` returns immediately and the first namespace-present event crashes the process outright.

Setting `NotifyStartedFunc` only on the branch that does `wg.Add(1)` keeps the pairing intact.

cmd/routedns/main.go
```diff
@@ -506,6 +510,7 @@ func run(opt options, args []string) error {
 		}
 		pl := pendingListener{id: id, nsName: nsName, build: build}
 		if pl.nsName == "" {
+			wg.Add(1)
```
folbricht, line 513: The `Add`/`Done` pairing also assumes each listener starts exactly once, but both restart paths call `Start()` repeatedly: the supervisor loop retries every second after an error, and `superviseNetNSListener` rebuilds and restarts on every namespace flap. `notifyStarted()` fires on each successful start, so a second successful bind calls `wg.Done()` on an already-zero counter and panics.

This is reachable for DoH, ODoH, admin, DoQ and DTLS, which construct a fresh server on each `Start()`. For example a DoH listener whose `ServeTLS` returns on an accept error is retried a second later, binds fine, and takes the daemon down. DNS and DoT are shielded only incidentally, because `dns.Server` returns "server already started" on re-entry.

Wrapping each listener's `wg.Done` in a `sync.Once` makes the signal one-shot per listener regardless of restarts.

adminlistener.go
```diff
@@ -98,6 +98,7 @@ func (s *AdminListener) startTCP() error {
 		return err
 	}
 	defer ln.Close()
+	s.opt.notifyStarted()
```
folbricht, line 101: `startQUIC` (line 106 onward) has no matching `notifyStarted()` call, so an admin listener with `transport = "quic"` increments the WaitGroup and never decrements it. The listener binds and serves normally, but `wg.Wait()` in main.go blocks forever, READY is never sent, systemd `Type=notify` kills the unit at `TimeoutStartSec`, and `Restart=on-failure` loops it.

`TestAdminListenerNotifyStartedFunc` only covers the default tcp transport, so this path is untested.

dohlistener.go
```diff
@@ -141,6 +141,7 @@ func (s *DoHListener) startTCP() error {
 		return err
 	}
 	defer ln.Close()
+	s.opt.notifyStarted()
```
folbricht, line 144: Same omission as the admin listener: `startQUIC` never calls `notifyStarted()`. `protocol = "doh"` with `transport = "quic"` is documented in doc/listeners.md, and it binds successfully but never signals readiness, hanging `wg.Wait()` permanently and failing the systemd start.

`TestDoHListenerNotifyStartedFunc` uses the default tcp transport and does not catch it.

cmd/routedns/main.go
```diff
@@ -551,6 +556,15 @@ func run(opt options, args []string) error {
 		go superviseNetNSListener(pl.id, pl.nsName, pl.build)
 	}
 
+	// Notify Systemd that we are done starting, i.e. the listeners are up.
+	// This is a no-op unless $NOTIFY_SOCKET is set.
+	wg.Wait()
```
folbricht, line 561: Two problems with waiting here.

The wait is unbounded. Any listener that cannot bind (port already in use, an address on an interface that is not up yet, a missing netns, either of the QUIC paths above) leaves the counter non-zero forever. Under the previous `Type=simple` the remaining listeners kept serving; now systemd kills the whole service at the start timeout.

The wait also sits before `signal.Notify` a few lines down, so while it is blocked no signal handler is registered and SIGTERM or SIGINT takes the Go default of immediate termination. The `onClose` handlers never run, so for instance the memory cache's write-to-file-on-close is skipped, all because an unrelated listener could not bind.

Registering the signal handler before the wait, and bounding the wait with a timeout that notifies READY anyway and logs a warning, decouples shutdown from startup completeness.

folbricht (commented):
Review of the readiness synchronization. Five correctness issues, three of which can panic or hang the daemon, plus three smaller points on the new tests.

dnslistener_test.go
```diff
@@ -0,0 +1,155 @@
...
+			}, new(TestResolver))
+			require.NoError(t, err)
+			return l
+		})
+}
+
+func TestDoQListenerNotifyStartedFunc(t *testing.T) {
+	testNotifyStartedFunc(t,
+		func(t *testing.T, f func()) Listener {
+			addr, err := getLnAddress()
```
folbricht, line 94: `getLnAddress()` reserves and releases a TCP ephemeral port, but this listener is UDP-based. `getUDPLnAddress()` in dotlistener_test.go is the UDP equivalent and is what doqlistener_test.go already uses.

The TCP-derived port may already be bound on UDP by another process, in which case the listener never starts and the test hangs to its full one-minute timeout. Same applies to the DTLS test at line 126 and the udp DNS test at line 45.

dnslistener_test.go
```diff
@@ -0,0 +1,155 @@
...
+func testNotifyStartedFunc(t *testing.T, makeListener func(*testing.T, func()) Listener) {
+	started := make(chan struct{})
+	notifier := &mockNotifier{}
+	notifier.On("NotifyStarted").Once().Return().Run(func(mock.Arguments) {
+		close(started)
+	})
+
+	l := makeListener(t, notifier.NotifyStarted)
+
+	go l.Start()
```
folbricht, line 31: The error from `Start()` is discarded. If it returns immediately on a bind failure or a bad TLS config, the test blocks the full minute and reports `timeout` rather than the real cause. Sending the error into a channel and selecting on it alongside `started` would make failures diagnosable.

dnslistener_test.go
```diff
@@ -0,0 +1,155 @@
...
+type mockNotifier struct {
+	mock.Mock
+}
+
+// Used to mock  a NotifyStartedFunc
+func (m *mockNotifier) NotifyStarted() {
+	m.Called()
+}
+
+// NotifyStartedFunc is invoked once the ODoH listener is up, which is used by
```
folbricht, line 20: The comment says "the ODoH listener", but this helper is protocol-agnostic and is used by all seven listener types.

folbricht:
It's still in progress but I  just wanted to leave some comments early in case they're helpful. I'll not have time to work on this for a week, so no rush on any of it.

cmd/routedns/main.go
bbjubjub2494: Regarding problem 1, I believe reporting full failure is the sensible design here: if I make a unit dependent on routedns at the systemd level, there should be a guarantee that routedns is listening on all expected ports, not just maybe one or maybe the other. As a system administrator, I should then make routedns depend on network-related stuff so that its own preconditions are filled. Alternatively, if I'm ok with routedns being silently half-broken I should make a `Type=Simple` unit. Observe that the default behavior of `dnsmasq` is to exit if it is not able to bind all specified sockets, so there's likely a user expectation here (although routedns doesn't immediately exit in this case)

Problem 2 is well-taken.

